PR #970, opened 2015-01-16
Remove old Wiki links and redirect to sensible replacement content.

juliandunn:
Also, it's bad to send people to the Chinese wiki :grin:

### DIFF
--- a/app/views/pages/dashboard.html.erb
+++ b/app/views/pages/dashboard.html.erb
@@ -80,7 +80,7 @@
 
     <h4>Resources</h4>
     <ul class="pretty">
-      <li><%= link_to 'Learn Chef', 'https://learnchef.com', target: 'blank' %></li>
+      <li><%= link_to 'Learn Chef', learn_chef_url, target: 'blank' %></li>
       <li><%= link_to 'Join IRC', 'irc://irc.freenode.net/chef' %></li>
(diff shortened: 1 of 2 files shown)


app/views/pages/dashboard.html.erb
```diff
@@ -80,9 +80,9 @@
 
     <h4>Resources</h4>
     <ul class="pretty">
+      <li><%= link_to 'Learn Chef', 'https://learnchef.com', target: 'blank' %></li>
```
smith: We have helpers you should be using for these. Use `learn_chef_url` instead of `'https://learnchef.com'`.

See https://github.com/opscode/supermarket/blob/master/app/helpers/custom_url_helper.rb.

app/views/pages/welcome.html.erb
```diff
@@ -51,7 +51,7 @@
       <h2>Share</h2>
       <ul>
         <li>
-          <a href="https://wiki.opscode.com/display/ChefCN/Creating+New+Cookbooks" target="_blank">Share your cookbooks</a>
+          <a href="http://docs.chef.io/knife_cookbook_site.html#share" target="_blank">Share your cookbooks</a>
```
smith: Use `"#{chef_docs_url}/knife_cookbook..."` instead.

juliandunn:
Hey @smith now that we killed the Wiki, could I get a re-review so we can get this in 1.6.0?

smith:
:+1:

PR #976, opened 2015-01-24
Remove frightening destruction message

juliandunn:
Change the unsubscription messages to not use the word "destroy", because in this context it sounds like Supermarket `rm -rf`'d the repo.


smith:
:+1:

Issue #975, opened 2015-01-24
add link somewhere to list of curry-managed repositories


juliandunn:
As a Supermarket admin,
I would like to see what repositories are being checked for a CLA,
So that I don't have to dig through HipChat history to remember that the link is https://supermarket.chef.io/curry/repositories

bcobb:
https://github.com/chef/supermarket/blob/7cba5b63f4d99c5d174c6d11b90314c59aec532a/app/views/application/_appheader.html.erb#L31-L35 makes me think that you should be seeing the link when you hover over your username in the header.

juliandunn:
You are right! But that label is confusing - I never clicked on it because I thought it was to show the cookbooks that I personally was following.

kirtfitzpatrick:
What should that label be?

juliandunn:
is this not already closed by #978?

PR #978, opened 2015-01-27
Clarify text in the link to Curry's repositories

bcobb:
:fork_and_knife: 

Here's my stab at addressing @juliandunn's comment in #975. Notably, it transitions from a noun-y link to a verb-y link, like the others in the menu.


smith:
:+1: thanks!

PR #1004, opened 2015-03-11
removing non-working vagrant set up instructions

nellshamrell:
Removes Vagrant dev setup instructions that haven't worked for some time.  The local dev setup instructions do work.


kirtfitzpatrick:
👌

smith:
We should remove the related files in bin and exec as well.

PR #1013, opened 2015-04-07
Update link to supermarket cookbook in README.md

tuahk:
Noticed that this link was giving 404 so changed it to point to opscode-cookbooks/supermarket instead.


nellshamrell:
Thank you!

PR #1014, opened 2015-04-08
correcting OWNERS file to reflect current ownership

nellshamrell:
Corrects the owners file so that Kirt and I are listed.


smith:
Note that this file predates the [Maintainers policy](https://github.com/chef/chef-rfc/blob/master/rfc030-maintenance-policy.md), which we should probably use instead, and replace this with a MAINTAINERS.md (or toml @thommay?) version.

Issue #964, opened 2015-01-14
Feature:  Serve from private S3 buckets


leftathome:
Longtime readers of my issues know that I'm trying to set up a reasonably enterprise-y private Supermarket.  At the moment I've got it serving cookbooks out of a private S3 bucket.  Amazon best practice says this is fine.  Only if I'm going to do it, I should be using signed/expiring URLs.

The way I've implemented this is that I'm configuring the S3 variables except for the CloudFront URL...

... and I've had to make one teeny tiny face-smashingly easy change in the CookbookVersions controller.

Instead of using Paperclip's url method in the redirect_to, I change it to expiring_url(15) ( == number of seconds the URL is valid before expiring).  I don't know if this would break if I fell back to local storage.

The other two ways I can think of to implement this would be:

* Use Rails to proxy S3 by downloading the tarball and send_data'ing it (with nginx/varnish-friendly headers) to the client.  The only plus I can think of to this implementation is if I have Supermarket cookbook consumers who are in a totally firewalled environment where they can't hit S3...
* CloudFront (but ... how to make this private?).

leftathome:
Oh, also, some additional reading on expiring URLs in Paperclip:

https://github.com/thoughtbot/paperclip/wiki/Restricting-Access-to-Objects-Stored-on-Amazon-S3

leftathome:
Update!  Apparently expiring_url handling in Paperclip v4.2.0 (as used in Supermarket 1.5.x releases) didn't take into account that Fog handles expiring URL generation differently.  I'm going to try upgrading to v4.2.1 on my internal fork and see if that resolves some issues I've been having where cookbook download URLs are throwing 403 errors.  It seems to be random per CookbookVersion (but consistent on the ones that are affected)...

leftathome:
Update update: Yeah, that didn't help at all.  I should check commit diffs before getting excited about reading release notes.

However!  I did find something that fixed my issue - I added a fog API call to connection.sync_clock.  I've suspected that the temporary URL being generated was being generated with an expiration date in the past.  This is either down to bad time zone handling or some really weird environment configuration somewhere in omnibus-supermarket or the supermarket app itself.  So I'm mentioning it here first.  (No one else on the Google-indexed Internet seems to be having this problem at the moment.)

For the record, I patched Paperclip::Storage::Fog#directory to run connection.sync_clock() before returning directory info.  It seemed like the right place, though I suppose one could argue it should be done during connection setup.

[...]

PR #1020, opened 2015-04-14
Temporary urls for private s3 storage

nellshamrell:
Fix for #964

### DIFF
--- a/app/controllers/cookbook_versions_controller.rb
+++ b/app/controllers/cookbook_versions_controller.rb
@@ -11,7 +11,8 @@ class CookbookVersionsController < ApplicationController
     Cookbook.increment_counter(:web_download_count, @cookbook.id)
     Supermarket::Metrics.increment('cookbook.downloads.web')
 
-    redirect_to @version.tarball.url
+    url = ENV['S3_PRIVATE_URLS_EXPIRE'].present? ? @version.tarball.expiring_url(ENV['S3_PRIVATE_URLS_EXPIRE']) : @version.tarball.url
+    redirect_to url
   end
 
   #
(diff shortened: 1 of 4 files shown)


jf647:
Is is possible to not lock the private buckets to setting the object ACL?

Our bucket configuration uses S3 bucket policies.  We cannot set the ACL of an object, and if you try the request is rejected with an access denied error.  We rely on policies to protect the objects.

nellshamrell:
Looking into that now.

nellshamrell:
Thanks for the feedback!

If I were to make setting the s3 permissions optional here:
```bash
    if ENV['S3_PRIVATE_URLS'].present?
      options = options.merge(
        s3_permissions: :private
      )
```
Perhaps by adding another environmental variable.
```bash
    if ENV['S3_PRIVATE_OBJECT_ACLS'].present?
      options = options.merge(
        s3_permissions: :private
      }
    end
```
If you were to leave the S3_PRIVATE_OBJECT_ACLS environmental variable as nil, and therefore the s3_permissions were not set so as not to conflict with your bucket policies, would this solve your issue?

You would still need to set the expire time for the url in the S3_PRIVATE_URLS_EXPIRE environmental variable (though I’ll probably rename it to something like S3_URLS_EXPIRE to be a bit clearer).  Then you would have expiring urls, but without setting the s3_permissions to private. 

Thank you!

jf647:
Yes, that combination of environment variables would work.

Thanks

nellshamrell:
Coding that up now :)

nellshamrell:
The code changes are complete and tested.  Once the Travis CI build turns green, this is ready for review then merging!

someara:
lgtm

cwebberOps:
@leftathome Would love your thoughts on this PR, for, ummmm, reasons.

PR #1023, opened 2015-04-15
Remove commas from platform names in seed.rb

tuahk:
I forgot commas to platform names in seed.rb. We probably don't want them there. Sorry about that :smile:


nellshamrell:
No prob!

Issue #980, opened 2015-01-29
Feature: Search for cookbooks based on supported platform


andrewelizondo:
Use case: As a new user I'd like to be able to access the supermarket and filter/search for cookbooks that are supported on the platform I'm managing. e.g. show me all of the cookbooks that claim to support windows

jtimberman:
This would be a great feature. Potential problem is that the metadata `supports` doesn't do anything else anywhere so many people don't use it. But hey maybe then they'd start!

tuahk:
I also think this would be a very useful feature so I'd liked to tackle this one.

PR #1028, opened 2015-04-21
Search based on supported platforms

tuahk:
This starts work on issue #980.

I thought that I make my work so far public, in a hope of some feedback and ideas on how it should look like.

Idea is to add some kind of search filter where you could choose which platforms you want to include in search result.

### DIFF
--- a/app/views/application/_search.html.erb
+++ b/app/views/application/_search.html.erb
@@ -1,25 +1,50 @@
 <div class="search_bar">
-  <%= form_tag @search[:path], class: 'search_form', method: :get do %>
-    <div class="search_toggle">
-      <a class="button expand large" data-dropdown="search-types" rel="toggle-search-types"><span><%= @search[:name] %></span> <i class="fa fa-chevron-down"></i></a>
+    <%= form_tag @search[:path], class: 'search_form', method: :get do %>
+      <div class="quick_search">
+        <div class="search_toggle">
+          <a class="button expand large" data-dropdown="search-types" rel="toggle-search-types"><span><%= @search[:name] %></span> <i class="fa fa-chevron-down"></i></a>
 
-      <ul id="search-types" data-dropdown-content class="f-dropdown">
-        <li><a href="#" data-url=<%= cookbooks_path %> rel="toggle-cookbook-search">Cookbooks</a></li>
-        <li><a href="#" data-url=<%= tools_path %> rel="toggle-tool-search">Tools</a></li>
-      </ul>
-    </div>
+          <ul id="search-types" data-dropdown-content class="f-dropdown">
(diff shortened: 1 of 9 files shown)


nellshamrell:
Looking good so far!

nellshamrell:
How I imagine this working:

On the main page a user sees when they visit supermarket.chef.io, they will continue to see the prominent search field.  Beneath that search field, they will a see a link to "Advanced Search".  

When they click this link, they will then be taken to a view where they can somehow indicate platforms they wish the cookbooks returned to support.  Maybe a list of check boxes?  Text fields?  Drop downs?

tuahk:
Ok, that sounds good. I'll try it with check boxes and see how it looks. There might be too many different supported platforms though.

tuahk:
That latest commit adds a drop-down window where users can select which platforms to include in search result. 

Currently it should look something like this when closed:

screenshot - 26 04 2015 - 22 03 09

And like this when opened:

screenshot - 26 04 2015 - 22 09 10

There are still some issues with it but I start working with them if you guys like where this is going.

app/views/application/_search.html.erb
```diff
@@ -1,25 +1,50 @@
...
+                <label>
+                  <%=check_box_tag 'platforms[]', platform, params[:platforms] ? params[:platforms].include?(platform) : false %>
+                  <%= platform %>
+                </label>
+            </div>
+          <% end %>
+        </div>
+
+        <div class= "advanced_textfield">
+          <%= search_field_tag 'platforms[]', nil, placeholder: "You can write platform name here, if it's not on the list!", class: 'platform_search_textfield' %>
```
smith: I don't think the comma or the exclamation point are necessary here.

smith:
@tuahk I like where this is going. Nice work so far!

tuahk:
I removed WIP from this PR.

There are still functionalities that I feel should be added. For example advanced search should probably let users define multiple platforms in text input and keep what was written there after submit. But unfortunately I'm going to be super busy for the next couple of weeks and probably won't have time to implement those. 

However, I think this would already be a helpful feature for the supermarket and I'll of course try to find the time to fix any problems you might notice.

nellshamrell:
Thanks so much!  Reviewing this now.

nellshamrell:
Overall this looks great!  I worked on refining the styling a bit, do you mind incorporating these changes?  After that I think it's good to merge :)

https://github.com/tuahk/supermarket/compare/search-by-platforms...nellshamrell:search-by-platforms

tuahk:
Great! Those changes made this look much better. Really liked that arrow :)

nellshamrell:
Awesome.  Merging!

nellshamrell:
This is now live!

https://supermarket.chef.io

PR #1033, opened 2015-04-28
Updating Announcements Banner

nellshamrell:
The Announcement banner is currently not active on public Supermarket.  After this is deployed, it needs to be activated through rails console.  

ROLLOUT.enable(:announcement)  (I think, will verify) needs to be run from console.


smith:
:+1:

Issue #1012, opened 2015-04-07
Custom image support in addition to Gravatar


double-z:
As a user I would like to be able to have the option of adding a custom image in place of Gravatar pic and disabling Gravatar support.

nellshamrell:
Is this due to security concerns with using Gravitar?  Or just wanting to be able to have a custom image?

stanchan:
This is primarily due to rendering issues when Gravitar is blocked through a proxy.

nellshamrell:
Labeling as BUG for the issues when Gravitar is blocked through a proxy and FEATURE for allowing custom images.  I'll work on the bug first :)

nellshamrell:
BUG part of this is complete, taking the Bug label off and returning this to the backlog.

nellshamrell:
We're not going to move forward with this at this time, but I greatly appreciate all the input!

PR #1038, opened 2015-05-05
Disable gravatar option

nellshamrell:
Fixes the bug part of #1012.  The feature aspect of it will be addressed in a later pull request.

When this is merged, a user can use ENV['DISABLE_GRAVATAR'] to disable calls to the external Gravatar service.


app/helpers/users_helper.rb
```diff
@@ -16,14 +16,7 @@ module UsersHelper
...
   def gravatar_for(user, options = {})
-    options = {
-      size: 48
-    }.merge(options)
-
-    size = options[:size]
-    gravatar_id = Digest::MD5.hexdigest(user.email.downcase)
-    gravatar_url = "https://secure.gravatar.com/avatar/#{gravatar_id}?s=#{size}"
-    image_tag(gravatar_url, alt: user.name, class: 'gravatar')
+    ENV['DISABLE_GRAVATAR'].present? ? no_gravatar_image : gravatar_image(user, options)
```
smith: Should we use Rollout with `FEATURES=gravatar` instead of an env var for this?

app/helpers/users_helper.rb
```diff
@@ -38,4 +31,21 @@ def pluralized_stats(count, thing)
...
+    }.merge(options)
+
+    size = options[:size]
+    gravatar_id = Digest::MD5.hexdigest(user.email.downcase)
+    gravatar_url = "https://secure.gravatar.com/avatar/#{gravatar_id}?s=#{size}"
+    image_tag(gravatar_url, alt: user.name, class: 'gravatar')
+  end
+
+  def no_gravatar_image
+    image_tag("apple-touch-icon.png", alt: "Default Avatar", class: 'gravatar')
```
smith: Can we use the `user.name` here as well for the `alt` attribute?

nellshamrell: That's a good idea!

app/helpers/users_helper.rb
nellshamrell: That's a good idea...though I'm not sure I full understand how Rollout works.  I know we define the rollout features in a databag in the community cookbook, where would a user define the rollout features in private Supermarket?

smith: They get set up by an environment variable: `FEATURES=a,b,c,d,...` In omnibus they get defined here: https://github.com/chef/omnibus-supermarket/blob/a1c159ce0793287b154a9edad1473ad75e762ed5/cookbooks/omnibus-supermarket/attributes/default.rb#L388-L403

nellshamrell: TY!

nellshamrell:
Follow on:  After this is deployed to public supermarket, we will need to activate the gravatar feature.  We will also need to convey this to customers who are using private versions of Supermarket.

spec/helpers/users_helper_spec.rb
```diff
@@ -1,15 +1,35 @@
...
+        ROLLOUT.deactivate(:gravatar)
+      end
     end
 
-    it "returns the image tag for the specified user's gravtar image with size" do
-      user = create(:user, email: 'johndoe@example.com')
-      expect(gravatar_for(user, size: 128)).to eq('<img alt="John Doe" class="gravatar" src="https://secure.gravatar.com/avatar/fd876f8cd6a58277fc664d47ea10ad19?s=128" />')
+    context 'when gravatar feature is disabled' do
+      before do
+        expect(ROLLOUT.active?(:gravatar)).to eq(false)
```
smith: We're not testing whether this feature is active are we? Wouldn't it make more sense to use something like:

```ruby
allow(ROLLOUT).to receive(:active?).with(:gravatar).and_return(false)
```

smith: Should we be doing the same thing with Rollout in both of these contexts? Either do the real activation and deactivation or stub the behavior. It looks like they are being treated differently.

nellshamrell: Agreed, fixing.

nellshamrell:
Fixed inconsistency in the specs!  

This is now ready for final review.

This is a potentially (but not severe) breaking update for customers running a private Supermarket, as they will need to activate the gravatar feature.  I propose the release this one is included in is released as 2.0.0-beta0

nellshamrell:
Any objection to merging?

PR #1039, opened 2015-05-05
Remove the Ruby 2.1.3 pin from the Gemfile, because we're already pinning it in omnibus-supermarket

irvingpop:
Also update `:github` lines in the Gemfile to use https instead.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,7 @@
 source 'https://rubygems.org'
 
-# TODO: is value needs to be updated as local dev changes, and removed when
-# local dev doesn't need to specify it anymore
-ruby ENV['CUSTOM_RUBY_VERSION'] || '2.1.3'
+# Ruby version isn't specified here anymore, as it's pinned in Omnibus
+# https://github.com/chef/omnibus-supermarket/blob/master/config/projects/supermarket.rb
 
 gem 'rails', '~> 4.1.5'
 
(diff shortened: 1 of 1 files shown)


smith:
We should pin the same version that we do in Omnibus, so that we run the same version in dev as in prod.

:+1: to changing this to be the same as what we're shipping in Omnibus, :-1: to removing it from here.

irvingpop:
hey @smith how about something like this, to satisfy both the Omnibus and your current local dev use cases:
```ruby
ruby ENV['CUSTOM_RUBY_VERSION'] || '2.1.3'
```

and maybe a `TODO` comment above it stating that this value needs to be updated as local dev changes, and removed when local dev doesn't need to specify it anymore?

smith:
> hey [@smith](https://github.com/smith) how about something like this, to satisfy both the Omnibus and your current local dev use cases:

> ruby ENV['CUSTOM_RUBY_VERSION'] || '2.1.3'
> and maybe a `TODO` comment above it stating that this value needs to be updated as local dev changes, and removed when local dev doesn't need to specify it anymore?

That sounds like it makes things more complex and I don't see what benefit it would have. Let's just remove it and link to the omnibus override line in the Gemfile and/or README.

irvingpop:
@smith done! :)

nellshamrell:
:+1:

PR #1040, opened 2015-05-06
Make search-bar as wide as the whole page when using xlarge screen

tuahk:
I noticed that my last PR #1028 had an unintentional side effect. When you use supermarket on a large screen it fails to keep search-bar as wide as the whole page. So here is a fix for that.


smith:
Thanks so much for the fix @tuahk!

PR #1043, opened 2015-05-09
stop rendering autodependencies

lamont-granquist:
cookbook autodependencies are bad to feed into depsolvers.  cookbooks
all automatically are dependencies of themselves, but some people still
do:

name 'foo'
depends 'foo'

what this can actually do is create depsolver loops where all the
versions of a cookbook create dependency loops which satsify this
constraint and make the depsolver work much harder than it should.

this patch just stops supermarket from publishing this information out
and starts to stop the pain.

additionally, there should be a migration added to clean these up and
there should be filtering on the input side.  knife and the chef server
also need to clean them up as well on the client and server side.

at the same time, though, we need to not completely break old cookbooks,
so we'll probably be stuck with maintaining cookbooks that have
auto dependencies in metadata.rb for some time, but we can gradually
make the tools all start ignoring those deps.

### DIFF
--- a/app/models/universe.rb
+++ b/app/models/universe.rb
@@ -39,6 +39,7 @@ module Universe
       FROM cookbook_versions
         INNER JOIN cookbooks ON cookbooks.id = cookbook_versions.cookbook_id
         LEFT JOIN cookbook_dependencies ON cookbook_dependencies.cookbook_version_id = cookbook_versions.id
+          AND cookbook_dependencies.name != cookbooks.name
     )
 
     cookbooks = ActiveRecord::Base.connection.execute(sql).to_a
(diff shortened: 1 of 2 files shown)


lamont-granquist:
paging @danielsdeleo and @manderson26 thoughts?

I've also got a PR into foodcritic to add a rule there.

Also looking at a PR to chef/knife so cookbook uploads and downloads will strip self-dependencies and will whine about them.

danielsdeleo:
LGTM

nellshamrell:
This looks pretty straight forward.

@lamont-granquist, once this is merged (or before) can you create Github issues for each of the follow up tasks?  i.e. the db migration and the GUI changes you'd like to see.

lamont-granquist:
Yeah, I was gonna just submit PRs as I knocked them down, but I can submit issues now in case I squirrel and forget about them.

nellshamrell:
This looks fine to me and is a good step.

PR #1044, opened 2015-05-09
add rake tasks to verify development/test env

miketheman:
`rake setup:deps` validates that the correct minimum versions are
installed and are running.

/cc: @nellshamrell @kirtfitzpatrick


nellshamrell:
This looks fine to me.  @kirtfitzpatrick, do you think so as well?

PR #1045, opened 2015-05-10
add zebra striping to adoption list

miketheman:
- Use CSS3 built-in method for selecting list pattern children
- Remove one level of `li` nesting

> As a Supermarket Viewer
When viewing the [Adoption Cookbooks page](https://supermarket.chef.io/available_for_adoption)
Then I want to be able to visually differentiate quickly between lines.

I tried to achieve this in two methods. This PR and another here: https://github.com/chef/supermarket/compare/chef:master...miketheman:scss_simple_listing

I think this method is the "cleaner" one, as it cleans up one level of LI nesting, and uses a [built-in selector](http://www.w3schools.com/cssref/sel_nth-child.asp) instead of using Rails logic to tag items when rendering.

The end result of both methods is the same, appears like so:
screen shot 2015-05-10 at 10 07 00 am

I wasn't certain which coloring to use, so I chose one that seemed reasonably viewable.

### DIFF
--- a/app/assets/stylesheets/simple_listing.scss
+++ b/app/assets/stylesheets/simple_listing.scss
@@ -14,6 +14,10 @@
     display: none;
   }
 
+  li:nth-child(odd) {
+    background-color: $secondary_gray;
+  }
+
   li {
     @include dark-links;
     border-bottom: rem-calc(2) solid lighten($secondary_gray, 32%);
--- a/app/assets/stylesheets/simple_listing.scss
+++ b/app/assets/stylesheets/simple_listing.scss
@@ -14,7 +14,7 @@
     display: none;
   }
 
-  li:nth-child(odd) {
+  &.zebra li:nth-child(odd) {
     background-color: $secondary_gray;
   }
 
--- a/app/views/cookbooks/available_for_adoption.html.erb
+++ b/app/views/cookbooks/available_for_adoption.html.erb
@@ -12,7 +12,7 @@
     </h1>
   </div>
 
-  <ul class="simple_listing">
+  <ul class="simple_listing zebra">
     <% @available_cookbooks.each do |cookbook| %>
       <%= render "simple_cookbook", cookbook: cookbook %>
     <% end %>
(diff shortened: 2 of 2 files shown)


miketheman:
The `li` nesting removal gets rid of these extra empty items that have no visible purpose, as they are 0 pixels high, and only increase the overall size of the html file.

screen shot 2015-05-10 at 11 09 32 am

If you wanted to leave the extra `li` in there, then `li:nth-child(odd)` should become `li:nth-child(4)` - but that is less clear to me overall.

nellshamrell:
This looks great, Mike, are you going to submit the other change as a pull request as well?  I'm not seeing it here.

miketheman:
@nellshamrell I referenced the other as another method to execute the same - one that I discovered while trying to accomplish this one. I left is as something to contemplate alternatives, if this changeset was undesirable.

nellshamrell:
Got it, ty!

nellshamrell:
Hi Mike,

This unfortunately affects not just the adoptable cookbooks list, but also the lists of featured cookbooks on the /cookbooks-directory page.  Can you isolate this to the adoptable cookbooks list?

screen shot 2015-05-11 at 3 25 26 pm

miketheman:
Sure! The other approach would have been isolated to the adoption page, so
I'll try to narrow the scope in CSS first, and if that doesn't work, use
that method.

nellshamrell:
Thanks!

miketheman:
Hi @nellshamrell,
I pushed a change that allows any unordered list to opt-in to using the `zebra` class selector - and there's currently only one! :grin:

miketheman:
For reference, the other implementation method would have had the same problem - the modification in that one was in the `_simple_cookbook.html.erb` template, which both pages use - whereas this modification is scoped to the page wrapping the simple_cookbook rendering.
So basically, controlling the `li` from outside the template.

**Cascading** Style Sheets, indeed! 
TMYK

nellshamrell:
Indeed!
css

nellshamrell:
Looks great!  Merging.

PR #1054, opened 2015-05-26
Make user admin task

nellshamrell:
This adds a rake task users can use to promote an existing user to Admin.

This is intended to be called by a supermarket-ctl command, which I will add to the [omnibus-supermarket](https://github.com/chef/omnibus-supermarket) repo.

I will update this when the pull request to supermarket-ctl is also ready to go.  Once that is added, this should be good to merge.

### DIFF
--- a/app/services/make_user_admin.rb
+++ b/app/services/make_user_admin.rb
@@ -7,10 +7,30 @@ class MakeUserAdmin
 
   def call
     begin
-      User.find(@user)
+      user = User.find(@user)
+      add_admin_role(user)
     rescue ActiveRecord::RecordNotFound
-      "User not found in Supermarket.  Make sure the user exists in Supermarket before making it an admin."
+      user_not_found_message
     end
+  end
+
+  private
+
+  def user_not_found_message
+    "User not found in Supermarket.  Make sure the user exists in Supermarket before making it an admin."
+  end
+
+  def successful_promotion_message(user)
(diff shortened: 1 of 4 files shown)


app/services/make_user_admin.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def call
+    user = User.with_username(@user_name).first
+    user.present? ? add_admin_role(user) : user_not_found_message
+  end
+
+  private
+
+  def user_not_found_message
+    'User not found in Supermarket.  Make sure the user exists in Supermarket before making it an admin.'
```
smith: Since we're pulling in the Rails environment in the Rake task, can we just use i18n tags for these instead?

smith:
> I will update this when the pull request to supermarket-ctl is also ready to go. Once that is added, this should be good to merge.

What kind of dependency does this have on supermarket-ctl? I don't understand why something else would be needed in order for this to be merged. Am I misunderstanding something?

nellshamrell:
It's not actually depended on supermarket-ctl, it can be merged without doing anything with omnibus-supermarket.  However, in order for a user to access this rake task through a supermarket-ctl command, this pull request will need to be merged as well.

https://github.com/chef/omnibus-supermarket/pull/22

app/services/make_user_admin.rb
nellshamrell: Good idea.  Implementing.

Issue #1046, opened 2015-05-11
Cookbook self-dependency cleanup checklist


lamont-granquist:
- [x] filter outgoing /universe endpoint to not render self-dependencies
- [ ] filter outgoing cookbook_versions to not render self-dependencies
- [x] filter incoming cookbook uploads so that stored metadata will not have self-dependencies
- [ ] add database migration to strip existing self-dependencies out of cookbooks

lamont-granquist:
The universe endpoint filtering will be handled by #1043 (which also has most of the background on this task in it)

lamont-granquist:
(I'm not positive that the second task is that useful or worthwhile, but included it for completeness -- it might actually be useful to still have self-dependencies show up in the GUI so that when knife complains about a self-dependency that user's can see the self-dependency in the GUI and not hide that from users the way that #1043 hides it from berkshelf).

nellshamrell:
@lamont-granquist: going to close this for now, but if you complete further work on it please feel free to re-open it!

PR #1055, opened 2015-05-26
part #2 of removing cookbook self-deps

lamont-granquist:
strip out self-dependencies on upload before the cookbook records are
written to the database.


lamont-granquist:
part of #1046

smith:
:+1: , but rubocop burned you for an unused block element: https://travis-ci.org/chef/supermarket/builds/64173544#L529-L531

Either fix that or put in the exception to tell rubocop we don't care.

nellshamrell:
LGTM

Issue #1062, opened 2015-06-10
Add ability to specify path for S3 buckets


nellshamrell:
Right now users can specify a S3 bucket for cookbook storage, but they cannot specify a path to store the bucket.  Paperclip does have the ability to specify a path for file storage, but we need to add in an environmental variable so that the user can set this in their supermarket install.

See http://stackoverflow.com/a/7476436 for a code example.

PR #1063, opened 2015-06-11
Ability to specify s3 bucket paths

nellshamrell:
Fix for #1062

### DIFF
--- a/config/initializers/paperclip.rb
+++ b/config/initializers/paperclip.rb
@@ -15,7 +15,7 @@ end
         access_key_id: ENV['S3_ACCESS_KEY_ID'],
         secret_access_key: ENV['S3_SECRET_ACCESS_KEY']
       },
-      path: path,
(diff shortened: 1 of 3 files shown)


config/initializers/paperclip.rb
```diff
@@ -15,7 +15,7 @@
         access_key_id: ENV['S3_ACCESS_KEY_ID'],
         secret_access_key: ENV['S3_SECRET_ACCESS_KEY']
       },
-      path: path,
+      path: "#{ENV['S3_PATH']}/#{path}",
```
kirtfitzpatrick: Is this a breaking change? What will happen to existing supermarket's that upgrade and don't specify 'S3_PATH'? Won't that add a preceding '/' to the path?

nellshamrell: That's an excellent point.  I'll add in a guard to that.

kirtfitzpatrick:
fixes #1063

PR #1070, opened 2015-06-29
Ensure we only set a maintainer once

thommay:


nellshamrell:
@thommay are you still working on this one?

thommay:
Can this wait for a while? I need to do some other things and then come back and keep going on this.

thommay:
@nellshamrell can we resurrect this one?

db/schema.rb
```diff
@@ -11,11 +11,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150512134228) do
+ActiveRecord::Schema.define(version: 20150626120818) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
-  enable_extension "pg_trgm"
```
nellshamrell: Any reason for removing this extension?

thommay: i assume rails did that when i ran `rake db:migrate`, because it certainly wasn't intentional.

smith: Only if you want to make search break :)

nellshamrell: Can you re-add it?  Otherwise this looks good.

nellshamrell:
:+1: Merging :)

PR #1072, opened 2015-06-30
add ability of admins to add collaborators

lamont-granquist:
This means that I can add myself to chef cookbooks as an admin and
then can upload cookbooks.  Otherwise I have to pester Sean even
though I'm an "admin", although I can issue a transfer request but
that is going to be even more annoying.


nellshamrell:
:+1:

PR #1106, opened 2015-08-19
Add missing protocol to links in Manage Profile

rafaelmagu:
Noticed a couple of broken links while trying to configure my machine to share cookbooks to public Supermarket.


nellshamrell:
Thank you!

PR #1131, opened 2015-09-24
Curry mass subscribe script

### LINKED ISSUE #1127: Curry needs to be updated for chef-cookbooks move
Our current curry config has all the cookbooks in opscode-cookbooks.  The Github api doesn't handle the move.  We need to update these all to chef-cookbooks.  Simple little change to enable curry.

nellshamrell:
This pull request is meant to provide a script to fix #1127 

What this does is grabs all the public repositories in the chef-cookbooks organization (which we recently removed all cookbooks that used to be in opscode-cookbooks into) and checks whether they are watched by Curry.  If they are, it does nothing. If they are not, it Curry subscribes to them to verify CLAs when someone submits a pull request.

As it's written right now, this is meant to be a one-off script.  I am open to adding it to the application, but it will need a little more finesse in that case.


robbkidd:
:+1:

Seems good to me as a newb grocer. I'm assuming it is intended to be run from the production rails console. As you say, if it is to be added to the application as more than a one-off, I suggest I've got some thoughts on finesse.

tas50:
:+1: Logic seems sound

nellshamrell:
Ran the script and it appears to have been successful.  Please let me know if you see further issues!

lib/supermarket/curry_mass_subscribe.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    private
+
+    def client
+      Octokit::Client.new(access_token: ENV['GITHUB_ACCESS_TOKEN'])
+    end
+
+    def public_repos(org)
+      client.org_repos(org, private: true)
```
nellshamrell: There might be something amiss with this line, even though the script seemed to run fine.  I feel like it should have said client.org_repos(org, private: false) to return public repos.  However, when I ran this line individually, it definitely picked up public repos (and returned the same number of repos regardless of whether we specified private being true or false), so there might be something incorrect with how we are calling the org repos and trying to filter them so that we only get the public ones.  I'm not merging this script into the code base, so I'm going to go ahead and close this pull request and leave the script as it is, but I will leave this note as a caution to anyone who might come across this pull request in the future.

Issue #1139, opened 2015-10-07
Review use of Time.now


robbkidd:
There are a few places that use `Time.now` instead of the [recommended](http://rails-bestpractices.com/posts/2014/10/22/use-time-zone-now-instead-of-time-now/) `Time.zone.now` or `Time.current`.

```
» ack Time.now app
app/models/ccla_signature.rb
35:  before_create -> (record) { record.signed_at ||= Time.now }

app/models/cla_report.rb
73:    report.csv_file_name = "#{Time.now.to_i}.csv"

app/models/cookbook.rb
270:      self.updated_at = Time.now

app/models/icla_signature.rb
31:  before_create -> (record) { record.signed_at ||= Time.now }

app/views/tools/index.atom.builder
3:  feed.updated Time.now
```

Recommend review and correction.

robbkidd:
Reviewed. Being worked in #1142

robbkidd:
_poke_

robbkidd:
_poke again_

PR #1142, opened 2015-10-09
Change all calls to Time.now into Time.current

robbkidd:
Per a few resources' recommendations[1], calls for current time in a Rails
app should be to `Time.current` not `Time.now`.

[1] Resources:
* http://rails-bestpractices.com/posts/2014/10/22/use-time-zone-now-instead-of-time-now/
* http://danilenko.org/2012/7/6/rails_timezones/
* http://winstonyw.com/2014/03/03/time_now_vs_time_zone_now/

Addresses issue #1139

### DIFF
--- a/lib/supermarket/health.rb
+++ b/lib/supermarket/health.rb
@@ -56,7 +56,7 @@ module Supermarket
       postgres_health_metric do
         @supermarket[:expired_ocid_tokens] = Account.
           for('chef_oauth2').
-          where('oauth_expires < ?', Time.now).
+          where('oauth_expires < ?', Time.current).
           count
       end
     end
--- a/app/models/ccla_signature.rb
+++ b/app/models/ccla_signature.rb
@@ -32,7 +32,7 @@ class CclaSignature < ActiveRecord::Base
 
   # Callbacks
   # --------------------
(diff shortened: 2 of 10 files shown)


robbkidd:
In general, I believe the use of `Time.now` was mitigated by [the application config default time zone being set to UTC](https://github.com/chef/supermarket/blob/master/config/application.rb#L68). The setting of values to `Time.current` should not affect things going forward. I'll comment on when a `Time.current` is being used for comparison.

lib/supermarket/health.rb
```diff
@@ -56,7 +56,7 @@ def expired_ocid_tokens
       postgres_health_metric do
         @supermarket[:expired_ocid_tokens] = Account.
           for('chef_oauth2').
-          where('oauth_expires < ?', Time.now).
+          where('oauth_expires < ?', Time.current).
```
robbkidd: Values coming out of the database will be in UTC. ActiveRecord auto-converts time to UTC (per the app config) so there should be no behavioral difference in use of `Time.now` vs `Time.current` in this query.

nellshamrell:
:+1:

nellshamrell:
TY for this!

PR #1146, opened 2015-10-22
Add collaborator refactor 2

nellshamrell:
This refactors the logic around adding a collaborator out of the collaborator controller into a concern.  I realized this refactor was needed during the process of coding the groups feature for Supermarket.  When that feature is launched, we will need the ability to add collaborators (and all the checking that is part of that process) to be accessible from more places than just one action in the collaborator controller.  Rather than repeat the code in multiple controllers, I refactored it into the concern you see here.


robbkidd:
Collaborator factory concern! :+1: :shipit:

PR #1147, opened 2015-10-27
refactors removing a collaborator logic into a shared controller concern

nellshamrell:
This is similar to #1146, but it refactors removing a collaborator into the shared controller concern.  This continues to lay the foundation for the groups feature, which requires the ability to remove a collaborator from multiple controllers.  The reason this is in a controller concern, rather than a service object, is that Pundit (our authorization service) requires access to controller methods such as current_user.

### DIFF
--- a/spec/controllers/concerns/collaborator_processing_spec.rb
+++ b/spec/controllers/concerns/collaborator_processing_spec.rb
@@ -121,7 +121,7 @@ describe FakesController do
 
     let!(:collaborator) { create(:cookbook_collaborator, resourceable: cookbook, user: hank) }
 
-    it 'deletes a collaborator if the signed in user is the resource owner' do
+    it 'allows a resource owner to remove a collaborator' do
       sign_in fanny
 
       expect do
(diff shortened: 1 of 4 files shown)


spec/controllers/concerns/collaborator_processing_spec.rb
```diff
@@ -114,4 +114,39 @@ class FakesController < ApplicationController
...
+      sign_in fanny
+
+      expect do
+        subject.remove_collaborator(collaborator)
+      end.to change { Collaborator.count }.by(-1)
+
+      expect(response).to be_success
+    end
+
+    it 'deletes a collaborator if the signed in user is a collaborator on this resource' do
```
robbkidd: The description of this spec is ambiguous to me. Is this scenario "a user can remove their own collaboration access" or "a collaborator can remove another collaborator's access"? Code implies the former, but I'm not sure of the intention.

nellshamrell: Will clear that up.

nellshamrell: Done

robbkidd:
:+1:

PR #1151, opened 2015-10-30
Groups Feature

### LINKED ISSUE #658: team support for managing cookbooks
I'd like to see teams able to be created and managed.

This will help us manage the users on more then one cookbook at a time (we plan on managing a few cookbooks) as people move around.

This would be for ongoing cookbook user management, just just a one time thing.

nellshamrell:
https://trello.com/b/SuBjc0us/supermarket-groups

Closes #658


.env
```diff
@@ -18,7 +18,7 @@ CURRY_SUCCESS_LABEL="Signed CLA"
 PUBSUBHUBBUB_SECRET=YOUR_PUBSUBHUBBUB_SECRET
 NEW_RELIC_APP_NAME=Supermarket
 PUBSUBHUBBUB_CALLBACK_URL=http://example.com
-FEATURES=cla,join_ccla,tools,fieri,announcement,github
+FEATURES=cla,join_ccla,tools,fieri,announcement,github, collaborator_groups
```
nellshamrell: adds collaborator groups to the default features for use in development environments

.env
```diff
@@ -17,7 +17,7 @@ CURRY_CLA_LOCATION=https://community.chef.io
 CURRY_SUCCESS_LABEL="Signed CLA"
 PUBSUBHUBBUB_SECRET=YOUR_PUBSUBHUBBUB_SECRET
 NEW_RELIC_APP_NAME=Supermarket
-FEATURES=cla,join_ccla,tools,fieri,announcement,github
+FEATURES=cla,join_ccla,tools,fieri,announcement,github, collaborator_groups
```
nellshamrell: Adds collaborator_groups to default list of features in .env file

app/assets/javascripts/application.js
```diff
@@ -27,6 +27,8 @@
 //= require flash
 //= require select2.min
 //= require collaborators
+//= require groups
+//= require group_members
```
nellshamrell: Require new javascript files related to groups

app/assets/javascripts/group_members.js
```diff
@@ -0,0 +1,33 @@
+$(document).on('opened', '#group-members[data-reveal]', function () {
```
nellshamrell: This is new javascript to autopopulate the text box when searching for a user to add to a group as a group member.

app/assets/javascripts/groups.js
```diff
@@ -0,0 +1,37 @@
+$(document).on('opened', '#collaborators[data-reveal]', function () {
```
nellshamrell: This is new javascript to autopopulate the groups field when adding a group as a collaborator.

app/controllers/collaborators_controller.rb
```diff
@@ -36,7 +36,9 @@ def create
         collaborator_params[:resourceable_id]
```
nellshamrell: Adds logic for groups to the current collaborators controller

app/controllers/concerns/collaborator_processing.rb
```diff
@@ -5,13 +5,14 @@ module CollaboratorProcessing
     helper_method :add_users_as_collaborators
```
nellshamrell: Adds groups logic to the collaborator_processing concern

app/controllers/group_members_controller.rb
```diff
@@ -0,0 +1,91 @@
+class GroupMembersController < ApplicationController
```
nellshamrell: New controller for group_members, which belong to groups!

app/controllers/groups_controller.rb
```diff
@@ -0,0 +1,52 @@
+class GroupsController < ApplicationController
```
nellshamrell: New groups controller

app/controllers/users_controller.rb
```diff
@@ -33,6 +33,15 @@ def tools
   end
 
   #
+  # GET /users/:id/groups
+  #
+  # Display a user and their groups
+  #
+  def groups
```
nellshamrell: simple method to see what groups a user belongs to

app/helpers/collaborators_helper.rb
```diff
@@ -0,0 +1,9 @@
+module CollaboratorsHelper
```
nellshamrell: Useful for views involving collaborators - distinguishes between collaborators that are associated with groups and those that are not

app/helpers/groups_helper.rb
```diff
@@ -0,0 +1,18 @@
+module GroupsHelper
+  def admin_member?(user, group)
```
nellshamrell: helper methods for groups views

app/models/collaborator.rb
```diff
@@ -3,16 +3,18 @@ class Collaborator < ActiveRecord::Base # Associations
   # --------------------
```
nellshamrell: adds groups logic to collaborator model

app/models/cookbook.rb
```diff
@@ -138,16 +139,21 @@ def sorted_cookbook_versions
   #
   def transfer_ownership(initiator, recipient)
     if initiator.is?(:admin) || collaborator_users.include?(recipient)
-      collaborator = Collaborator.new(user_id: user_id, resourceable: self)
       update_attribute(:user_id, recipient.id)
-      collaborator.save!
+
+      unless collaborators.where(user_id: initiator.id).present?
```
nellshamrell: Handle the unlikely but possible scenario when a group may have a member who is NOT a collaborator on a resource the group is associated with.  This is also prevented by logic on the controller...but I want to ensure that we have error handling in case the scenario is caused by a rails console change or something to that effect.

app/models/cookbook.rb
```diff
@@ -138,16 +139,21 @@ def sorted_cookbook_versions
...
 
       if collaborator_users.include?(recipient)
         collaborator = collaborators.where(
           user_id: recipient.id,
           resourceable: self
         ).first
-        collaborator.destroy unless collaborator.nil?
+        # Do not destroy collaborator is collaborator does not exist
+        # OR if the collaborator is associated with a group
+        collaborator.destroy unless collaborator.nil? || collaborator.group_id.present?
```
nellshamrell: Do not destroy a collaborator during an ownership transfer if the collaborator is associated with a group

app/models/group_resource.rb
```diff
@@ -0,0 +1,7 @@
+class GroupResource < ActiveRecord::Base
```
nellshamrell: GroupResource is the join table that joins groups with resources (cookbooks or tools).

app/views/collaborators/_add.html.erb
```diff
@@ -3,10 +3,15 @@
...
 <%= form_for Collaborator.new do |f| %>
   <div class="row collapse">
-    <div class="small-10 columns">
+    <div class="small-12 columns">
       <%= f.hidden_field :resourceable_type, value: resource.class.name %>
       <%= f.hidden_field :resourceable_id, value: resource.id %>
-      <%= f.hidden_field :user_ids, class: 'collaborators multiple', 'data-url' => collaborators_path(ineligible_user_ids: Collaborator.ineligible_collaborators_for(resource).map(&:id)) %>
+      <%= f.label :users %>
+      <%= f.hidden_field :user_ids, id: 'user_ids', class: 'collaborators multiple', 'data-url' => collaborators_path(ineligible_user_ids: Collaborator.ineligible_collaborators_for(resource).map(&:id)) %>
+      <% if ROLLOUT.active?(:collaborator_groups) %>
```
nellshamrell: Adds a groups field to the add a collaborator form.  Only shows the field if the collaborators_groups feature is active.

app/views/group_members/_add.html.erb
```diff
@@ -0,0 +1,7 @@
+<h1>Add Group Members</h1>
```
nellshamrell: Dialog to add a new user to a group as a group member.

app/views/groups/_group.html.erb
```diff
@@ -0,0 +1,8 @@
+<li itemscope itemtype="http://schema.org/Code">
```
nellshamrell: This is for a list of groups that a user will see when they click the "Groups" button on their profile page.  This lists all groups the user is a member of.

app/views/groups/_member.html.erb
```diff
@@ -0,0 +1,21 @@
+<li itemscope itemtype="http://schema.org/Code">
+  <div class="header">
+    <div class="header-content">
+      <%= link_to member.user, rel: 'member', itemprop: 'username' do %><%= member.user.username %><% end %>
+      <div class="remove">
+        <% if admin_member?(current_user, member.group) %>
+          <% if member.admin? %>
+            <% if admin_members(member.group).count > 1 %>
```
nellshamrell: Only show remove button for an admin member when there is more than one admin member of the group.

app/views/groups/index.json.jbuilder
```diff
@@ -0,0 +1 @@
+json.groups @groups, :id, :name
```
nellshamrell: Respond to json requests asking for a list of groups (see the GroupsController#index page for more info)

app/views/resources/_collaborators.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="collaborators_avatar">
```
nellshamrell: Render collaborators associated with a group and collaborators NOT associated with a group in separate places of the view.

app/views/users/_heading.html.erb
```diff
@@ -1,10 +1,13 @@
...
 
-  <% if ROLLOUT.active?(:tools) %>
     <div class="buttons">
       <%= link_to "Cookbooks", user, class: "button radius secondary #{'active' if active == 'cookbooks'}", rel: "view_cookbooks" %>
-      <%= link_to "Tools & Plugins", tools_user_path(user), class: "button radius secondary #{'active' if active == 'tools'}", rel: "view_tools" %>
-    </div>
+  <% if ROLLOUT.active?(:tools) %>
+    <%= link_to "Tools & Plugins", tools_user_path(user), class: "button radius secondary #{'active' if active == 'tools'}", rel: "view_tools" %>
   <% end %>
+  <% if ROLLOUT.active?(:collaborator_groups) %>
```
nellshamrell: Only show groups link if the collaborator_groups feature is active

prometheanfire:
squash please

.env
robbkidd: That extra space, tho!

nellshamrell: Fixed!

app/controllers/collaborators_controller.rb
```diff
@@ -58,6 +60,38 @@ def destroy
     end
   end
 
+  # DELETE /collaborators/:id/transfer
```
robbkidd: According to `rake routes`, I believe this action is `DELETE /collaborators/:id/destroy_group`

```
destroy_group_collaborator DELETE /collaborators/:id/destroy_group(.:format) collaborators#destroy_group
```

app/controllers/group_members_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+          group_id: group_member_params[:group_id]
+        )
+        if group_member.save
+          group_resources(group_member).each do |resource|
+            add_users_as_collaborators(resource, group_member.user.id.to_s, group_member.group.id)
+          end
+        else
+          flash[:warning] = group_member.errors.full_messages.join(', ')
+          redirect_to group_path(group_member_params[:group_id])
+          return
```
robbkidd: Given that we're iterating over a collection of `user_ids` and saving an object for each, suggest we shovel errors onto the `flash[:warning]` to collect any that fail and allow the iteration to continue.

app/controllers/concerns/collaborator_processing.rb
```diff
@@ -23,11 +24,24 @@ def add_users_as_collaborators(resource, user_ids)
...
+
   def remove_collaborator(collaborator)
     authorize!(collaborator, 'destroy?')
     collaborator.destroy
   end
 
+  def remove_group_collaborators(collaborators)
+    collaborators.each do |collaborator|
+      remove_collaborator(collaborator)
+    end
```
robbkidd: We could consider having this method return a collection of users whose `Collaboration` instances were destroyed as a result of removing a `Group`'s collaboration access to a resource.

Method signature might look like `remove_group_collaborators(resource, group)` and have this method perform the `Collaborator` lookup logic currently in the `collaborators_controller`.

```
def group_collaborators(resource, group)
  Collaborator.where(resourceable: resource, group: group)
end
```

app/controllers/group_members_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+  end
+
+  def destroy
+    if @group_member.destroy
+
+      group_resources(@group_member).each do |resource|
+        collaborator = resource.collaborators.where(user_id: @group_member.user_id).first
+        remove_collaborator(collaborator) if collaborator.present?
+      end
```
robbkidd: This seems to be another iteration that might have failures we should collect. Possibly wrap the whole `@group_member.destroy` and iterative `remove_collaborator` in a transaction?

app/models/cookbook.rb
```diff
@@ -407,4 +401,23 @@ def deprecate_search(query)
   def copy_name_to_lowercase_name
     self.lowercase_name = name.to_s.downcase
   end
+
+  def create_new_collaborator(initiator)
+    unless collaborators.where(user_id: initiator.id).present?
```
robbkidd: And `where(group_id: nil)`? This method will not create a new `Collaborator` if the `initiator` is a member of a collaborating group. That might be intentional, just checking.

app/models/cookbook.rb
```diff
@@ -407,4 +401,23 @@ def deprecate_search(query)
...
+  def delete_old_collaborator(recipient)
+    if collaborator_users.include?(recipient)
+      collaborator = collaborators.where(
+        user_id: recipient.id,
+        resourceable: self
+      ).first
+      # Do not destroy collaborator is collaborator does not exist
+      # OR if the collaborator is associated with a group
+      collaborator.destroy unless collaborator.nil? || collaborator.group_id.present?
+    end
```
robbkidd: Based on the `unless`, it looks like we want to find a non-group-affiliated `Collaborator` for the `recipient` user and delete that one. More tinkering:

```
collaborator = collaborators.where(
  user_id: recipient.id,
  group_id: nil
).first
collaborator.destroy if collaborator
```

Also, in staring at these methods, I note that from the private methods' perspectives, they do not care that the parameter happens to be an `initiator` or a `recipient`. The param is just a `user`. The context of what kind of user only matters up in the body of the `transfer_ownership` method.

app/models/cookbook.rb
```diff
@@ -407,4 +401,23 @@ def deprecate_search(query)
   def copy_name_to_lowercase_name
     self.lowercase_name = name.to_s.downcase
   end
+
+  def create_new_collaborator(initiator)
+    unless collaborators.where(user_id: initiator.id).present?
+      collaborator = Collaborator.new(user_id: initiator.id, resourceable: self)
+      collaborator.save!
+    end
```
robbkidd: Removed my previous comment after reading deeper in the spec and seeing that we do intend to only create a `Collaborator` for the `initiator` [if that user does not already have collaborator access via a group membership](https://github.com/nellshamrell/supermarket/blob/groups-feature-2/spec/models/cookbook_spec.rb#L209-L227).

I'll did some tinkering and had tests pass with `collaborators.where(user_id: initiator.id).first_or_create!`.

app/views/collaborators/_add.html.erb
```diff
@@ -3,10 +3,15 @@
 
 <%= form_for Collaborator.new do |f| %>
   <div class="row collapse">
-    <div class="small-10 columns">
+    <div class="small-12 columns">
       <%= f.hidden_field :resourceable_type, value: resource.class.name %>
       <%= f.hidden_field :resourceable_id, value: resource.id %>
-      <%= f.hidden_field :user_ids, class: 'collaborators multiple', 'data-url' => collaborators_path(ineligible_user_ids: Collaborator.ineligible_collaborators_for(resource).map(&:id)) %>
+      <%= f.label :users %>
+      <%= f.hidden_field :user_ids, id: 'user_ids', class: 'collaborators multiple', 'data-url' => collaborators_path(ineligible_user_ids: Collaborator.ineligible_collaborators_for(resource).map(&:id)) %>
```
robbkidd: As a note for future improvement, we should consider moving AR queries out of views and into controllers.

app/views/groups/_member.html.erb
robbkidd: This was a bit gnarly to follow. What do you think about condensing the "show the remove button" logic to a single `if`?

```erb
<% if admin_member?(current_user, member.group) %>
        
  <% if !member.admin? || (member.admin? && admin_members(member.group).count > 1) %>
    <%= link_to "Remove", group_member_path(member.id), method: :delete, data: { confirm: "Are you sure you want to remove this member?" }, class: 'button tiny alert radius' %>
  <% end %>

  <% if !member.admin? %>
    <%= link_to "Make Admin", make_admin_group_member_path(member.id), method: :post, class: 'button tiny radius' %>
  <% end %>

<% end %>
```

```ruby
!member.admin? || (member.admin? && admin_members(member.group).count > 1 )`
```
... could become ...

```ruby
!member.admin? || admin_members(member.group).count > 1
```

... but then we might be back in "what's going on here?" territory again.

robbkidd: Also noticed `@admin_members` in the `GroupsController` that can be used at least for the count instead of a query to the database through `admin_members` method from the helper.

```
admin_member?(current_user, member.group)
``` 
... might be able to be replaced with ...

```
@admin_members.map(&:id).include?(current_user.id)
```

... maybe? To avoid more trips to the database? Maybe in the helper?

app/views/resources/_collaborators.html.erb
```diff
@@ -1,5 +1,5 @@
...
-  <% collaborators.each do |collaborator| %>
+  <% non_group_collaborators(collaborators).each do |collaborator| %>
```
robbkidd: I wish I could come up with a better name for these folks.

app/views/resources/_collaborators.html.erb
```diff
@@ -13,4 +13,14 @@
       <% end %>
     </div>
   <% end %>
+
+  <% if resource.group_resources.present?  && ROLLOUT.active?(:collaborator_groups) %>
```
robbkidd: Maybe flip the order of these for a fast fail when the feature is not enabled. No trip to the DB to look up non-existent `group_resources` when the feature is turned off.

app/controllers/collaborators_controller.rb
```diff
@@ -58,6 +60,38 @@ def destroy
...
+  # id is group's id
+  #
+  # Removes a group of collaborators
+  #
+  def destroy_group
+    group = Group.find(params[:id])
+    resource = params[:resourceable_type].constantize.find(
```
robbkidd: We'll need to guard this call to `constantize` on something coming from user input.

spec/controllers/collaborators_controller_spec.rb
```diff
@@ -88,6 +87,96 @@
...
+        let(:group_resources) { GroupResource.where(group: group, resourceable: cookbook) }
+
+        let(:collaborators) { Collaborator.where(group: group, resourceable: cookbook) }
+
+        before do
+          sign_in fanny
+        end
+
+        it 'finds the correct group' do
+          expect(Group).to receive(:find).and_return(group)
```
robbkidd: For this to test the correctness of the `find`, I think it needs a `.with(group.id)` included in the expectation. In the current form, the expectation acts more like a stubbed `Group.find` method.

spec/controllers/concerns/collaborator_processing_spec.rb
```diff
@@ -115,7 +115,93 @@ class FakesController < ApplicationController
...
+        subject.add_group_members_as_collaborators(cookbook, "#{group.id}")
+      end
+
+      it 'associates the collaborator with the group' do
+        expect(cookbook.collaborators).to be_empty
+        subject.add_group_members_as_collaborators(cookbook, "#{group.id}")
+        cookbook.collaborators.each do |collaborator|
+          expect(collaborator.group).to eq(group)
+        end
+      end
```
robbkidd: I suggest that we only keep these last two specs in this context to avoid specifying too much of the implementation in the specs.

* `associates the collaborator with the group` confirms that the new `Collaborator` objects are members of the test group—arguably the most important test.
* `makes a collaborator for each group user` tests that `add_users_as_collaborators` is called the correct number of times with the correct signature.
* Maybe add a third spec for testing `associate_group_to_resource` gets called with the correct signature.

Then we can rely on the tests for `add_users_as_collaborators` and `associate_group_to_resource` to be more detailed.

spec/controllers/concerns/collaborator_processing_spec.rb
```diff
@@ -115,7 +115,93 @@ class FakesController < ApplicationController
...
+
+    context 'adding multiple groups' do
+      let!(:group2_member) { create(:group_member) }
+      let(:group2) { group2_member.group }
+      let!(:group2_member2) { create(:group_member, group: group2) }
+
+      it 'finds all groups' do
+        expect(Group).to receive(:find).at_least(:twice).and_return(group, group2)
+        subject.add_group_members_as_collaborators(cookbook, "#{group.id}, #{group2.id}")
+      end
```
robbkidd: Suggest tweaking this to be similar to revised specs for the "adding a single group" context:

* two specs for testing that `add_users_as_collaborators` and `associate_group_to_resource` get called the correct number of times with correct signatures
* one spec—currently present in `makes a new collaborator for each user in both groups`—to test like a black box

app/controllers/collaborators_controller.rb
nellshamrell: Fixored!

app/views/groups/_member.html.erb
nellshamrell: Done!

app/controllers/collaborators_controller.rb
nellshamrell: Fixored!

app/controllers/group_members_controller.rb
nellshamrell: Done!

app/models/cookbook.rb
nellshamrell: Good thinking, done!

app/views/resources/_collaborators.html.erb
nellshamrell: :+1:

app/controllers/concerns/collaborator_processing.rb
nellshamrell: Going back and forth on whether this would make things clearer or more confusing to troubleshoot...

app/controllers/group_members_controller.rb
nellshamrell: Good thinking...though I think we're ok because we are removing only one group member at a time, rather than all the group members within a group.  Additionally, we do verify that a collaborator is present before attempting to remove the collaborator from the resource.  I can see both sides...but I'd like to leave it as it is for now.

app/models/cookbook.rb
nellshamrell: Good idea, done!

spec/controllers/concerns/collaborator_processing_spec.rb
nellshamrell: Good thinking...but I think I'd rather leave it as it is now.  Thanks for the suggestions, though!

spec/controllers/concerns/collaborator_processing_spec.rb
nellshamrell: Good thinking...but I think I'd rather leave it as it is now.  Thanks for the suggestions, though!

robbkidd:
:+1:

Issue #1049, opened 2015-05-13
Incremental update of out of date gems


nellshamrell:
@irvingpop wisely pointed out that we have some out of date gems in use in Supermarket.  Although we first tried to take the "fix them all at once" approach, we kept seeing spec failures building with each change.  I 100% agree we should update the gems, but I'd like to take more of an incremental approach to lessen the risk of breaking things.

Please refer to @irvingpop's pull request #1042  for guidelines on which gems should be updated, then I'd like to take the approach of updating three or four at a time, getting specs passing, deploying, then building on it with the next gem updates.

irvingpop:
+1

It would be really cool to run `bundle-audit` in CI runs, once we get to a good state.

nellshamrell:
:+1:

nellshamrell:
Concentrate on security issue gems first

nellshamrell:
This was partially solved by #1075, moving back to the backlog for the time being.

nellshamrell:
Definition of Success: We will call this one closed with bundle-audit shows no security risk gems.

robbkidd:
```
~/supermarket ±master [8d2a69d] » be bundle-audit
No vulnerabilities found
```

So we've got that going for us.

robbkidd:
It helps if you update bundle-audit's security vulnerability database. PR incoming to address most of the vulnerabilities reported at Gemnasium.

robbkidd:
Vulnerable gems updated. We still have some that are "out-of-date," but I don't currently consider that a bad thing.

I propose expanding the definition of done given by @nellshamrell above ("We will call this one closed with bundle-audit shows no security risk gems.") to include running `bundle-audit` as part of the test suite.

robbkidd:
Closing now that bundler-audit reports we're clean.

PR #1153, opened 2015-11-02
Update gems with security notices

robbkidd:
Gem updates to address security issues
* `shoulda-matchers` (to gem version, remove GitHub pinning since gem version now  includes the change in the commit to which we were pinned)
* `uglifier` had an issue (that doesn't affect current Supermarket, but upgrading because green is better than red on the vulnerability scanner)
* `chef` gem appears to be included to do version number validation, updated to a version that does not have conflicting RSpec version constraints. Upgrading RSpec is something that can be done later.

**UPDATED**

* `sidekiq` upgraded to version with fixes for vulnerabilities in the web UI (that supermarket does not use out-of-the-box)


robbkidd:
Security updates for #1049 . That leaves Sidekiq security issues:

```
» be bundle-audit
Name: sidekiq
Version: 3.2.3
Advisory: 125675
Criticality: Unknown
URL: https://github.com/mperham/sidekiq/pull/2422
Title: Sidekiq Gem for Ruby Multiple Unspecified CSRF
Solution: upgrade to >= 3.4.2

Name: sidekiq
Version: 3.2.3
Advisory: 125676
Criticality: Unknown
URL: https://github.com/mperham/sidekiq/issues/2330
Title: Sidekiq Gem for Ruby web/views/queue.erb CurrentMessagesInQueue Element
Reflected XSS
Solution: upgrade to >= 3.4.0

Name: sidekiq
Version: 3.2.3
Advisory: 125678
Criticality: Unknown
URL: https://github.com/mperham/sidekiq/pull/2309
Title: Sidekiq Gem for Ruby web/views/queue.erb msg.display_class Element XSS
Solution: upgrade to >= 3.4.0
```

These don't affect Supermarket out of the box because we don't run the Sidekiq web interface.

robbkidd:
```
» be bundle-audit update
Updating ruby-advisory-db ...
From https://github.com/rubysec/ruby-advisory-db
 * branch            master     -> FETCH_HEAD
Already up-to-date.
ruby-advisory-db: 233 advisories

» be bundle-audit
No vulnerabilities found
```

Gemfile
```diff
@@ -17,7 +17,7 @@ gem 'foreman'
 gem 'pundit'
 gem 'dotenv'
 gem 'octokit', require: false
-gem 'sidekiq'
+gem 'sidekiq', '= 3.4.2' # pinned to version prior to celluloid upgrade that brakes sidetiq
```
tas50: breaks

smith:
:+1: 

Not sure on the preference of the rest of the team, but I find it much easier to work off of a branch of the repo than a fork, since if I want to check it out on my machine I have to add a remote from your fork and check that branch out, rather than just checking out a branch.

robbkidd:
@smith Thanks! Merging.

On the process tangent, Nell and I have been working in forks for two reasons:

1. it's the way community contributors work without a commit bit
2. a fork is an additional guard against unintended pushes to branches in main repo

To your point, though, I agree there is an advantage to working in branches in the main repo: they're easier to get to—both for humans and for automatons. Reason No 2 above is probably mitigated by the git v2.0+ `config push.default` being `simple` instead of `matching` these days.

PR #1154, opened 2015-11-05
Add bundle-audit to CI test suite

robbkidd:
Pin bundler-audit to a version with `check --update` option. This
unreleased version lets us perform an update of the vulnerability
database during the check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,5 +27,6 @@ rvm:
 script:
   - bundle exec rake spec
   - bundle exec rake spec:rubocop
+  - bundle exec bundle-audit check --update
 services:
   - redis-server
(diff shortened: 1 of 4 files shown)


robbkidd:
Pondering whether a single `bundle exec rake spec:all` would be better or worse than each of the commands called by the tasks defined under `spec:all` being run individually by Travis.

robbkidd:
Once (if) this is closed, I propose closing #1049 .

robbkidd:
@irvingpop Following up on your suggestion to add `bundle-audit` to Supermarket's CI tests.

irvingpop:
:+1: super big fan of this!

robbkidd:
Wunderbar! I'm deputizing you as a maintainer, so four-eyeballs requirement met. :+1: :shipit:

PR #1160, opened 2015-11-12
Add detail from exception thrown during version validation

robbkidd:
The default message was vague for an invalid dependency or platform version
constraint. For example:

```
Stove::Cli | >>>> The Chef Server did not understand the request because it was malformed.
{ "error_code"=>"INVALID_DATA",
  "error_messages"=>["Version constraint is not a valid Chef version constraint"]
}
```

The error above did not contain enough information for a cookbook author to determine which of the versions they had declared in metadata was invalid.

This change includes the exception class and detail from the original exception thrown about what is invalid about the data given. With this change, the error above would appear as:

```
Stove::Cli | >>>> The Chef Server did not understand the request because it was malformed.
{ "error_code"=>"INVALID_DATA",
  "error_messages"=>["Version constraint is not a valid Chef version constraint. Chef::Exceptions::InvalidCookbookVersion: '2' does not match 'x.y.z' or 'x.y'"]}
```

Updated spec expectations to use blocks to call out the action under test and to check for a regex match among the validation errors instead of an exact string inclusion.

### DIFF
--- a/lib/active_model/validations/chef_version_constraint_validator.rb
+++ b/lib/active_model/validations/chef_version_constraint_validator.rb
@@ -34,8 +34,9 @@ module ActiveModel
       #
       def validate_each(record, attribute, value)
         Chef::VersionConstraint.new(value)
-      rescue Chef::Exceptions::InvalidVersionConstraint, Chef::Exceptions::InvalidCookbookVersion
-        record.errors.add(attribute, options.fetch(:message), value: value)
+      rescue Chef::Exceptions::InvalidVersionConstraint, Chef::Exceptions::InvalidCookbookVersion => e
+        msg = "#{options.fetch(:message)}. #{e.class}: #{e.message}"
+        record.errors.add(attribute, msg, value: value)
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lib/active_model/validations/chef_version_constraint_validator.rb
```diff
@@ -34,8 +34,9 @@ def initialize(options)
       #
       def validate_each(record, attribute, value)
         Chef::VersionConstraint.new(value)
-      rescue Chef::Exceptions::InvalidVersionConstraint, Chef::Exceptions::InvalidCookbookVersion
-        record.errors.add(attribute, options.fetch(:message), value: value)
+      rescue Chef::Exceptions::InvalidVersionConstraint, Chef::Exceptions::InvalidCookbookVersion => e
+        msg = "#{options.fetch(:message)}. #{e.class}: #{e.message}"
```
smith: Are there any cases where `options[:message]` can be empty? If so maybe we should pass a 2nd argument to fetch with a generic message so the message doesn't start with a period. If not, then never mind.

smith:
:+1:

lib/active_model/validations/chef_version_constraint_validator.rb
robbkidd: Good question. There should not be any case where `options[:message]` is nil at this point. A fetch-with-fallback is [provided in the `initialize()`](https://github.com/chef/supermarket/blob/robb/add-detail-to-ver-constraint-errors/lib/active_model/validations/chef_version_constraint_validator.rb#L20-L22) for this validator, so a default value is present already at the point this validation method is called.

PR #1163, opened 2015-11-20
Use container infrastructure in travis

tas50:
Use the apt add-on to remove the need for sudo
Use use the actual libxslt package vs. the virtual that isn't approved
by travis


nellshamrell:
LGTM :+1:

robbkidd:
:+1:

PR #1188, opened 2015-12-21
Guard 3rd party calls in test

### LINKED ISSUE #1187: Wrap some specs in VCR
There are signs that some specs are making calls to 3rd party services. Find them and wrap them in VCR tape.

robbkidd:
With some judicious application of firewall rules, hunted down the places where tests were still making calls to 3rd parties.

* Wrapped the search engine ping for the sitemap update in a VCR cassette.
* Wrapped the header CDN inclusions such that they are not loaded in the test environment

Fixes #1187


robbkidd:
@nellshamrell I believe this is ready to merge.

nellshamrell:
:+1:

Issue #1156, opened 2015-11-05
Send longer changelog e-mails to cookbook followers


tas50:
The changelog e-mails we send are too short.  For our community cookbooks the user only gets a single line of update and even that is cropped. Here's a nice example of a confusing e-mail:

Changelog
memcached Cookbook CHANGELOG
This file is used to list changes made in each version of the memcached cookbook.

1.9.0 (2015-11-05)
NOTE: This will be the last version of this cookbook

nellshamrell:
Thanks @tas50!  Can you give us an example of an email that would be more useful to you?  I would love to have a template to start with :)

tas50:
In general it just needs to be a bit longer.

nellshamrell:
I understand, but can you give me an example of content that would be useful to include?

Thanks!

nellshamrell:
@tas50 - pinging you on this again, I would love to make it longer, but need an example of what content it is lacking and should be included to make the email longer.

tas50:
Sorry for the late response.  It's a bit tricky since you'd really have to parse the changelog and that would assume it follows a standard format.  I think there's 2 small things we could fix to improve it though.

Here's an example from an e-mail:

```Changelog
homebrew Cookbook CHANGELOG
This file is used to list changes made in each version of the homebrew cookbook.

v2.0.0 (2015-12-01)
Removed all Chef 10 compatibility code
#77 Update the tap
```

The very last line is cut off mid line.  Instead of including a certain number of characters we should include a certain number of lines.  Additionally at the end we should include a line with ... or some other marker to indicate there's additional content if you click the "View Full Changelog" button.

Here's the relevant changes for this version from the changelog

```
homebrew Cookbook CHANGELOG
This file is used to list changes made in each version of the homebrew cookbook.

v2.0.0 (2015-12-01)
Removed all Chef 10 compatibility code
#77 Update the tap provider to properly notify on changes
#73 Allow specifying versions (or HEAD) of formulas (see readme for usage)
Updated contributing, testing, and maintainers docs
Updated contents of chefignore and .gitignore files
Updated development dependencies in the Gemfile
Added Travis CI and supermarket version badges to the readme
Added Chef standard rubocop file and resolved all warnings
Added super metadata for Supermarket
Added testing in Travis CI
#75 Fix Chefspecs to properly run on Linux hosts (like Travis)
Add Rakefile for simplified testing
Resolved all foodcritic warnings
```

[...]

PR #1189, opened 2015-12-22
Longer Changelog in Cookbook Follower emails

nellshamrell:
Fix for #1156


robbkidd:
:+1: Yay for simplest change.

PR #1196, opened 2015-12-30
Add Compliance Profile as a tool type

nathenharvey:
Could it really be this simple?

I think it is based on 8f9fe9c and 846fc73

### DIFF
--- a/app/models/tool.rb
+++ b/app/models/tool.rb
@@ -1,7 +1,7 @@
 class Tool < ActiveRecord::Base
   include PgSearch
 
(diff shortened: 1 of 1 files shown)


robbkidd:
Behavior tested in development. Acts just like any other tool. At the moment, I'm inclined to merge this without specific testing around "compliance profile" as a tool type because (1) if tools are ducks, it acts like a duck and (2) we have some ideas about profiles becoming a fuller artifact type within supermarket in the future that will have their own non-tool-duck-like behavior.

robbkidd:
:+1:

PR #1197, opened 2015-12-31
adds distinctive view code for mobile browsers

### LINKED ISSUE #1069: Mobile view does not contain cookbook names
The Supermarket main page displays fairly well on mobile browsers, but for some reason the names of cookbooks are not displayed.  This is on an iPhone 6 with iOS 8.3.  See the attached screenshot.
img_0023

nellshamrell:
Fixes #1069

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -39,4 +39,15 @@ module ApplicationHelper
       'warning' => 'warning'
     }.fetch(name)
   end
+
+  #
+  # Returns true if request comes from a mobile device, false if not
+  #
+  # @param user_agent [String]
+  #
+  # @return [Bool]
+  #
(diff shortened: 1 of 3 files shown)


robbkidd:
After some tinkering in Chrome's mobile browser developer tools, I've seen that the disappearing names problem occurs in a desktop browser when the window width is shrunk. I believe the disappearing names are a result of [a CSS rule that hides A links](https://github.com/chef/supermarket/blob/master/app/assets/stylesheets/dashboard.scss#L28-L32) in the dashboard when screen width is under a certain size. I think the stylesheet would be a better place for a change than branching logic in the views.

nellshamrell:
Good point, will look into that.

nellshamrell:
Moved changes to https://github.com/chef/supermarket/pull/1205

PR #1202, opened 2016-01-12
Compliance Proflie tool type has shipped

nathenharvey:
Update the ROADMAP for Q1 2016


robbkidd:
:+1: Yes. Yes, it has.

PR #1206, opened 2016-01-20
Add a migration to ensure current DB extensions are enabled.

robbkidd:
These extensions are already included in the schema. But [pg_trgm was added directly to schema.rb](https://github.com/chef/supermarket/commit/99d7990d85e36821c57d43d2861316f46c0407de) leaving newly migrated databases (e.g. dev and test) without the extensions being included in the resultant post-migration schema.rb.

plpgsql is included from the PG adapter on platforms that need it, so it is not included directly in this migration.

pg_trgrm is needed for PG search to work.


nellshamrell:
:shipit:

PR #1207, opened 2016-01-20
Add local dev env things to gitignore

robbkidd:
* Mac's finder cache
* SASS cache
* direnv's .envrc and project directory when using `layout ruby`


nellshamrell:
:shipit:

PR #1212, opened 2016-01-26
Upgrade to Rails 4.1.14.1

robbkidd:
... to address multiple CVEs.


robbkidd:
@nellshamrell :eyes:

I'm not proud of the monkeypatch, but it gets us over the hump to be able to run the patched version of Rails. I added a naggy message to rspec output reminding us to get rid of that when we can upgrade the rspec-rails gem.

I tweaked some specs to be "test for 404 status in response" rather than "test for the 404 template to have been rendered" to get around some trouble between the latest Rails and the rspec-rails we're using. We can put those back later if we conclude that we like the render test better.

TrevorBramble:
@robbkidd: This looks great to me. Canny work with the RSpec workaround. :+1:

smith:
:+1:

robbkidd:
Excellent. Thanks @TrevorBramble and @smith. Imma squash these commits to clean up commentary.

robbkidd:
Commits squashed.

nellshamrell:
:+1:

PR #1214, opened 2016-01-27
Increase number of individual contributors shown in list.

robbkidd:


robbkidd:
@nellshamrell :eyes: This change will increase the number of _individual_ contributors shown on the page.

@tas50 :grimacing: :blush:

nellshamrell:
:shipit:

PR #1215, opened 2016-01-28
Upgrade rspec-rails, remove the need for a separate spec_feature_helper

robbkidd:
Upgrade to rspec-rails 3.4.1 removes the monkeypatch!

Moved the Rspec.configure bits from the feature helper into spec_helper.
They were conflicting under Rspec 3.2+ when features were run with other
spec types.

Made configs and includes for features feature specific in their declarations
instead of declaring them global but in the feature helper.

Updated README to reflect new spec landscape:
* don't need spec_feature_helper any more
* suite works with PhantomJS 2 now


robbkidd:
NO MONKEYPATCHES!

no capes

robbkidd:
And at least for this commit's Travis build, it cut the spec runtime in half.

nellshamrell:
I heartily support this!  :shipit:

PR #1218, opened 2016-02-02
Fix display of dependencies for a cookbook version other than latest

### LINKED ISSUE #1159: Dependencies tab of a cookbook appears to only show for current version
Using the rsyslog cookbook as an example of this bug.  If you look at version 3.0.0 of rsyslog, you'll see it is dependent on compat_resources.  If you use the pull down menu to switch to 2.2.0, it will continue to say it is dependent on compat_resources (even on page reload) - but in reality the source code for the 2.2.0 branch does not specify a dependency on compat_resources.

https://supermarket.chef.io/cookbooks/rsyslog/versions/2.2.0#dependencies

This leads to a confusing user experience.

robbkidd:
Fixes #1159

* Update development seeds to include 2 cookbook versions
* Even versions will depend on apt cookbook, odds on yum
* Give supported platforms a different version per version of cookbook, too

- [x] reduce duplication in the cookbook_view feature spec

### DIFF
--- a/app/views/cookbooks/_main.html.erb
+++ b/app/views/cookbooks/_main.html.erb
@@ -70,10 +70,10 @@
     </div>
     <div class="content" id="dependencies">
       <h3 class="dependent-header">Dependent cookbooks</h3>
-      <% if cookbook.cookbook_dependencies.present? %>
+      <% if version.cookbook_dependencies.present? %>
         <table>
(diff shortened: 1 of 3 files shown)


robbkidd:
out

nellshamrell:
Beautiful! :shipit:

Issue #1227, opened 2016-02-19
Make Fieri usable with a Private Supermarket


nellshamrell:
Fieri is currently a black box.  It is a separate service that runs Foodcritic specs on every cookbook that is uploaded to Supermarket.  Unfortunately, in its current state, private Supermarkets (especially ones that store their cookbooks on disk, rather than in an S3 bucket) cannot use it.  Some changes need to be made to Supermarket in order to make it able to use Fieri when it keeps cookbooks on disk.  This checklist will be updated as I go along.

- [x] Fix cookbook_version.tarball.url for Supermarkets that keep cookbooks on disk
- [x] Correct typo'd cookbook_version evaluate route (it's currently cookbook_verisons, two letters off)
- [x] Evaluate whether to refactor/correct the nginx config in Supermarket omnibus to redirect request for cookbooks when Supermarket keeps them on disk

PR #1230, opened 2016-02-23
Use the correct tarball url when Supermarket stores cookbooks on disk

nellshamrell:
Adds code to submit the correct tarball url to fieri when a Supermarket stores cookbooks on the local disk rather than S3

Fix for the first checklist item in #1227

### DIFF
--- a/app/workers/fieri_notify_worker.rb
+++ b/app/workers/fieri_notify_worker.rb
@@ -21,9 +21,25 @@ class FieriNotifyWorker
     data = {
       'cookbook_name' => cookbook_version.name,
       'cookbook_version' => cookbook_version.version,
-      'cookbook_artifact_url' => cookbook_version.tarball.url
+      'cookbook_artifact_url' => cookbook_artifact_url(cookbook_version)
     }
 
     response = Net::HTTP.post_form(uri, data)
+  end
+
+  private
+
+  def cookbook_artifact_url(cookbook_version)
+    if s3_configured?
+      cookbook_version.tarball.url
+    else
+      "#{Supermarket::Host.full_url}#{cookbook_version.tarball.url}"
+    end
+  end
+
+  def s3_configured?
+    %w(S3_BUCKET S3_ACCESS_KEY_ID S3_SECRET_ACCESS_KEY).all? do |key|
+      ENV[key].present?
+    end
   end
 end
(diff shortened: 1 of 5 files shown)


nellshamrell:
NOTE: I added a correction for the cookbook-versions#evaluation route.  I originally planned to have the old route (cookbook_verisons#evaluation - notice that the "i" and the "s" are transposed) kept for a few months...but it looks like you cannot have multiple routes direct to the same controller action.  Rails complains about it.  

When this is deployed, we MUST change a value in the fieri-staging and fieri-production data bags.

In the fieri-staging item within the apps databag, we need to change this:
results_endpoint:     https://supermarket-staging.getchef.com/api/v1/cookbook-verisons/evaluation

To this:
results_endpoint:     https://supermarket-staging.getchef.com/api/v1/cookbook-versions/evaluation

In the fieri-production item within the apps databag, we need to change this:
results_endpoint:     https://supermarket.getchef.com/api/v1/cookbook-verisons/evaluation

To this:
results_endpoint:     https://supermarket.getchef.com/api/v1/cookbook-versions/evaluation

Fieri will not work until this change is made after deploying this fix.

So this is officially a fix for #1227

app/workers/fieri_notify_worker.rb
```diff
@@ -21,9 +21,25 @@ def perform(cookbook_version_id)
...
-      'cookbook_artifact_url' => cookbook_version.tarball.url
+      'cookbook_artifact_url' => cookbook_artifact_url(cookbook_version)
     }
 
     response = Net::HTTP.post_form(uri, data)
   end
+
+  private
+
+  def cookbook_artifact_url(cookbook_version)
```
robbkidd: Does this method also need to handle [the case of S3 with expiring URLs](https://github.com/chef/supermarket/blob/master/app/controllers/cookbook_versions_controller.rb#L37-L43)? I only just found that. Some cursory reading on Paperclip doesn't surface a good abstraction for the attachment. :disappointed:

nellshamrell: Good question.

I don't *think* so, because you can only have all expiring urls with S3 or no expiring urls.  But let me verify.

nellshamrell: Oh...actually, it might.  Testing.

nellshamrell:
There is still work to be done in testing this and it is blocking #1254.  Closing this until the work on #1254 is complete.

PR #1242, opened 2016-03-02
Update steps and links to create/login Chef account to get a key.

### LINKED ISSUE #1241: Out of date instructions for user to get key from Chef Server
Need to update the text of https://github.com/chef/supermarket/blob/master/app/views/profile/edit.html.erb#L151 from `https://www.chef.io/account/password` to `https://id.chef.io/id/profile` 

`https://www.chef.io/account/password` was the old URL on the Chef Server, but now returns a 500 when hit.

robbkidd:
Fixes #1241.

### DIFF
--- a/app/views/profile/edit.html.erb
+++ b/app/views/profile/edit.html.erb
@@ -147,8 +147,9 @@
             can reset your key here:
           </p>
           <ol>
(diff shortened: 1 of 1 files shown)


nellshamrell:
:shipit:

PR #1245, opened 2016-03-04
Allow cookbook deprecations without a replacement cookbook

### LINKED ISSUE #1223: Add the ability to deprecate cookbooks without replacements
Some cookbooks just die and there's no replacement for them.  They were either bad ideas or the technology they automated is no longer around.  We need the ability to deprecate those without specifying a replacement cookbook.

robbkidd:
* Turn off JS to allow form to be submitted without content in the text
  field
* Turn the requirement for a cookbook->replacement parameter into a permit
  so that form can be submitted without a name
* Move the replacement cookbook lookup into the Cookbook.deprecate method
* Update deprecation messaging to not assume there is a replacement
  cookbook

Closes #1223

### DIFF
--- a/app/assets/javascripts/cookbookDeprecate.js
+++ b/app/assets/javascripts/cookbookDeprecate.js
@@ -28,8 +28,4 @@ $(document).on('opened', '[data-reveal]', function () {
   }
 
   $('.cookbook-deprecate').select2(settings);
-
-  $('.cookbook-deprecate').on("select2-selecting", function(e) {
-    $('.submit-deprecation').prop('disabled', false);
-  });
 });
--- a/app/controllers/cookbooks_controller.rb
+++ b/app/controllers/cookbooks_controller.rb
@@ -165,19 +165,16 @@ class CookbooksController < ApplicationController
   def deprecate
     authorize! @cookbook
 
-    replacement_cookbook = Cookbook.with_name(
-      cookbook_deprecation_params[:replacement]
-    ).first!
+    replacement_cookbook_name = cookbook_deprecation_params[:cookbook][:replacement]
 
-    if @cookbook.deprecate(replacement_cookbook)
+    if @cookbook.deprecate(replacement_cookbook_name)
       CookbookDeprecatedNotifier.perform_async(@cookbook.id)
 
       redirect_to(
         @cookbook,
         notice: t(
           'cookbook.deprecated',
-          cookbook: @cookbook.name,
-          replacement_cookbook: replacement_cookbook.name
+          cookbook: @cookbook.name
         )
       )
     else
@@ -281,7 +278,7 @@ class CookbooksController < ApplicationController
   end
 
   def cookbook_deprecation_params
-    params.require(:cookbook).permit(:replacement)
+    params.permit(cookbook: [:replacement])
   end
 
   def render_follow_button
--- a/app/mailers/cookbook_mailer.rb
+++ b/app/mailers/cookbook_mailer.rb
@@ -48,10 +48,10 @@ class CookbookMailer < ActionMailer::Base
     @email_preference = user.email_preference_for('Cookbook deprecated')
     @to = user.email
 
-    subject = %(
-      The #{@cookbook.name} cookbook has been deprecated in favor
-      of the #{@replacement_cookbook.name} cookbook
-    ).squish
+    subject = "The #{@cookbook.name} cookbook has been deprecated"
+    if @replacement_cookbook
+      subject += "in favor of the #{@replacement_cookbook.name} cookbook"
+    end
 
     mail(to: @to, subject: subject)
   end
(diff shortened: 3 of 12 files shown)


app/assets/javascripts/cookbookDeprecate.js
```diff
@@ -28,8 +28,4 @@ $(document).on('opened', '[data-reveal]', function () {
   }
 
   $('.cookbook-deprecate').select2(settings);
-
-  $('.cookbook-deprecate').on("select2-selecting", function(e) {
-    $('.submit-deprecation').prop('disabled', false);
-  });
```
robbkidd: Remove the now-unnecessary JS to enable the "Deprecate" submit button when something is selected in the text box.

app/controllers/cookbooks_controller.rb
```diff
@@ -165,19 +165,16 @@ def unfollow
   def deprecate
     authorize! @cookbook
 
-    replacement_cookbook = Cookbook.with_name(
-      cookbook_deprecation_params[:replacement]
-    ).first!
+    replacement_cookbook_name = cookbook_deprecation_params[:cookbook][:replacement]
```
robbkidd: Moved the replacement cookbook query into `Cookbook.deprecate` so we only need the name string here in the controller.

app/controllers/cookbooks_controller.rb
```diff
@@ -281,7 +278,7 @@ def cookbook_urls_params
   end
 
   def cookbook_deprecation_params
-    params.require(:cookbook).permit(:replacement)
+    params.permit(cookbook: [:replacement])
```
robbkidd: Changes the cookbook->replacement parameter from required to optionally-permitted.

app/models/cookbook.rb
```diff
@@ -351,21 +350,23 @@ def download_count
...
+  # @param replacement_cookbook_name [String] the name of the cookbook to
+  #   succeed this cookbook once deprecated
   #
   # @return [Boolean] whether or not the cookbook was successfully deprecated
   #   and  saved
   #
-  def deprecate(replacement_cookbook)
-    if replacement_cookbook.deprecated?
+  def deprecate(replacement_cookbook_name = '')
+    replacement_cookbook = Cookbook.find_by_name replacement_cookbook_name
```
robbkidd: Opted for `find_by_name` even though it does not do `with_name`'s lowercase searching. `find_by_name` returns the first match or a nil for us. Odds are that the incoming `replacement_cookbook_name` will match the existing cookbook because the web form was populated by a search of the database.

app/views/cookbooks/_manage_cookbook.html.erb
```diff
@@ -61,7 +61,7 @@
             <%= f.hidden_field :replacement, class: 'cookbook-deprecate', 'data-url' => deprecate_search_cookbook_path(cookbook) %>
           </div>
           <div class="small-3 columns">
-            <%= f.submit 'Deprecate', class: 'button radius postfix submit-deprecation', disabled: true %>
+            <%= f.submit 'Deprecate', class: 'button radius postfix submit-deprecation' %>
```
robbkidd: Related to the JS removal, this leaves the "Deprecate" button enabled for the form to submit empty text.

config/locales/en.yml
```diff
@@ -34,7 +34,7 @@ en:
       email_sent: "Ownership transfer email sent to %{user}."
       invite_accepted: "You are the new owner of %{cookbook}."
       invite_declined: "You have declined ownership of %{cookbook}."
-    deprecated: "%{cookbook} deprecated in favor of %{replacement_cookbook}."
+    deprecated: "%{cookbook} deprecated."
```
robbkidd: Removed the replacement from the internationalized success message for the flash. If there is a replacement, it already appears in the deprecation banner in the view.

nellshamrell:
:shipit:

app/mailers/cookbook_mailer.rb
```diff
@@ -48,10 +48,10 @@ def cookbook_deprecated_email(cookbook, replacement_cookbook, user)
     @email_preference = user.email_preference_for('Cookbook deprecated')
     @to = user.email
 
-    subject = %(
-      The #{@cookbook.name} cookbook has been deprecated in favor
-      of the #{@replacement_cookbook.name} cookbook
-    ).squish
+    subject = "The #{@cookbook.name} cookbook has been deprecated"
+    if @replacement_cookbook
+      subject += "in favor of the #{@replacement_cookbook.name} cookbook"
```
robbkidd: I need to add a space there. Whoops.

app/models/cookbook.rb
```diff
@@ -351,21 +350,23 @@ def download_count
...
       errors.add(:base, I18n.t('cookbook.deprecate_with_deprecated_failure'))
       return false
-    else
-      self.deprecated = true
-      self.replacement = replacement_cookbook
-      save
     end
+
+    self.deprecated = true
+    self.replacement = replacement_cookbook if replacement_cookbook.present?
```
robbkidd: This could reduce down to

```
self.replacement = replacement_cookbook.presence
```

Is that more or less readable with the trailing conditional removed?

PR #1252, opened 2016-03-24
Change the deprecation form text so that it's clear replacement is optional

robbkidd:
screen shot 2016-03-24 at 5 06 16 pm


robbkidd:
@nathenharvey :eyes:

tas50:
:+1:

PR #1253, opened 2016-03-25
Revised the copy of deprecation form and display banners

robbkidd:
Reads better and doesn't require the cookbook author to say all this in a
README update in a new cookbook release (even though that's still a good
thing to do.)

### modal form for declaring a cookbook is deprecated
screen shot 2016-03-25 at 3 46 08 pm

### banner for deprecated cookbook without a replacement
screen shot 2016-03-25 at 3 16 46 pm

### banner for deprecated cookbook with a recommended replacement
screen shot 2016-03-25 at 3 10 05 pm

:wink:


nathenharvey:
:+1:

PR #1261, opened 2016-04-05
Update paperclip gem to fix file upload DoS

robbkidd:
It's possible to cause a DoS by uploading files with a spoofed media type,
because it causes megabytes of logging to be written.


nellshamrell:
:+1: once tests pass

PR #1268, opened 2016-04-13
Add license info to project and software definitions

robbkidd:
Makes the omnibus packager happier to have licensing info for the project.


nellshamrell:
:+1:

PR #1269, opened 2016-04-13
Fix seeds with new location of README

robbkidd:
The README stayed at the root of the repo, so the seeds file needs the new path to it for faker cookbooks in development.


nellshamrell:
:shipit:

PR #1270, opened 2016-04-13
Fieri use with on disk cookbooks

nellshamrell:
This alters the cookbook_version.tarball.url to point to the correct location when a Supermarket is storing its cookbook on disc.

This also corrects a typo in one of our routes "cookbook_verisons" (switched i and s) is now "cookbook_versions."  I have left the old route in with a note in case anyone is still using it.  I have verified that both urls work with a live Private Supermarket and Fieri instance.

### DIFF
--- a/src/supermarket/app/workers/fieri_notify_worker.rb
+++ b/src/supermarket/app/workers/fieri_notify_worker.rb
@@ -21,23 +21,9 @@ class FieriNotifyWorker
     data = {
       'cookbook_name' => cookbook_version.name,
       'cookbook_version' => cookbook_version.version,
-      'cookbook_artifact_url' => cookbook_artifact_url(cookbook_version)
+      'cookbook_artifact_url' => cookbook_version.cookbook_artifact_url
     }
 
     response = Net::HTTP.post_form(uri, data)
-  end
-
- def cookbook_artifact_url(cookbook_version)
-    if s3_configured?
-      cookbook_version.tarball.url
-    else
-      "#{Supermarket::Host.full_url}#{cookbook_version.tarball.url}"
-    end
-  end
-
-  def s3_configured?
-    %w(S3_BUCKET S3_ACCESS_KEY_ID S3_SECRET_ACCESS_KEY).all? do |key|
-      ENV[key].present?
-    end
   end
 end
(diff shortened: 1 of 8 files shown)


nellshamrell:
This still needs to be tested with expiring S3 urls, will get on that tomorrow.

src/supermarket/app/workers/fieri_notify_worker.rb
```diff
@@ -21,9 +21,23 @@ def perform(cookbook_version_id)
...
     response = Net::HTTP.post_form(uri, data)
   end
+
+ def cookbook_artifact_url(cookbook_version)
+    if s3_configured?
+      cookbook_version.tarball.url
+    else
+      "#{Supermarket::Host.full_url}#{cookbook_version.tarball.url}"
+    end
+  end
```
robbkidd: Is this some logic we can push into the `Cookbook` model as an instance method? Seems like this worker knows too much.

src/supermarket/app/workers/fieri_notify_worker.rb
```diff
@@ -21,9 +21,23 @@ def perform(cookbook_version_id)
...
+
+  def s3_configured?
+    %w(S3_BUCKET S3_ACCESS_KEY_ID S3_SECRET_ACCESS_KEY).all? do |key|
+      ENV[key].present?
```
robbkidd: I _think_ we can ask `Paperclip` for [its storage option](https://github.com/chef/supermarket/blob/master/src/supermarket/config/initializers/paperclip.rb#L6-L16) somehow so that we only rely on the environment variables during initialization.

src/supermarket/app/workers/fieri_notify_worker.rb
nellshamrell: Good point!

src/supermarket/app/workers/fieri_notify_worker.rb
nellshamrell: Will investigate :)

robbkidd: I stumbled upon `Paperclip::Attachement.default_options[:storage]` while mucking about in the console.

src/supermarket/app/workers/fieri_notify_worker.rb
nellshamrell: Going to push to the CookbookVersion model, as urls are unique for each version.

nellshamrell:
This does indeed work with expiring S3 URLS when the expiration time is set to 30 seconds.  This is ready for review.

robbkidd:
👍

PR #1273, opened 2016-04-14
Add opensuseleap support

tas50:
Use the standard suse icon. Also while we're here add opensuse to search

### DIFF
--- a/src/supermarket/app/views/application/_search.html.erb
+++ b/src/supermarket/app/views/application/_search.html.erb
@@ -46,7 +46,7 @@
     <h4> Select Supported Platforms </h4>
     <div class = "advanced_search_platforms" >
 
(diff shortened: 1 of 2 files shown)


nellshamrell:
LGTM

src/supermarket/app/views/application/_search.html.erb
```diff
@@ -46,7 +46,7 @@
     <h4> Select Supported Platforms </h4>
     <div class = "advanced_search_platforms" >
 
-      <% %w(aix amazon centos debian fedora freebsd gentoo mac_os_x omnios openbsd oracle redhat ubuntu scientific smartos solaris suse windows zlinux).each do |platform| %>
+      <% %w(aix amazon centos debian fedora freebsd gentoo mac_os_x omnios openbsd opensuse opensuseleap oracle redhat ubuntu scientific smartos solaris suse windows zlinux).each do |platform| %>
```
robbkidd: I feel like this list could be centralized somewhere, but that's for another time.

PR #1275, opened 2016-04-14
Remove image editor information from the SVGs

tas50:
We could minimize our images a lot more, but this one is pretty obvious.  Remove the advertising for Illustrator / Sketch from the SVGs to shrink them all a tiny bit.


robbkidd:
👍  Images look right and slightly smaller.

PR #1278, opened 2016-04-18
adds ubuntu 12.04 to the list of kitchens

nellshamrell:
I've been manually adding Ubuntu 12.04 to the .kitchen.yml and see no reason not to include it in the .kitchen.yml stored in source code


robbkidd:
👍 Yes, please!

PR #1281, opened 2016-04-20
Make s3 location agnostic

### LINKED ISSUE #1266: Make region/url for S3 bucket configurable
Current behavior:
When a user specifies S3 storage for their cookbooks, the cookbook_version.tarball.url method defaults to "https://s3.amazonaws.com/bucket_name"

Problem:
The problem is this style of URL ONLY works when the S3 bucket is in N. Virginia, not any other regions.  To use any other region, the url must be "bucket_name.s3.amazonaws.com" OR "http://s3-aws-region.amazonaws.com/bucket".  This is likely configurable through Paperclip, we should add an environmental variable or something to that effect to configure the S3 region/url.

Desired behavior:
No matter what region a user's S3 bucket is in, cookbook_version.tarball.url uses the correct URL

nellshamrell:
Fixes #1266

### DIFF
--- a/src/supermarket/config/initializers/paperclip.rb
+++ b/src/supermarket/config/initializers/paperclip.rb
@@ -37,7 +37,7 @@ end
       )
     else
       options = options.merge(
-        url: ':s3_path_url'
+        url: ':s3_domain_url'
       )
     end
 


nellshamrell:
Confirmed that this works with both N.Virginia and other S3 locations.

robbkidd:
👍  :shipit: 🚀 🍰

PR #1286, opened 2016-04-22
"Recently Updated" cookbooks list now refers to list of recently shipped new versions

### LINKED ISSUE #1244: "Recently Updated" cookbooks list should only list recently shipped
Right now if you edit a cookbooks metadata on supermarket it shows up as being recently updated at https://supermarket.chef.io/cookbooks-directory

It seems like what people would really care about on this page is new cookbooks that have shipped.  We should only be showing a cookbook as being updated when the author ships a new version.  Adding issues URL or something like that shouldn't trigger it to the front page for everyone to see.

loganmeetsworld:
Rather than any change in the cookbook itself. 

Fixes #1244

### DIFF
--- a/src/supermarket/app/controllers/cookbooks_controller.rb
+++ b/src/supermarket/app/controllers/cookbooks_controller.rb
@@ -54,9 +54,8 @@ class CookbooksController < ApplicationController
   # Return the three most recently updated and created cookbooks.
   #
   def directory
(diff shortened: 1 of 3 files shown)


robbkidd:
👍  Pending making Rubocop happy.

robbkidd:
👍

nellshamrell:
👍

PR #1290, opened 2016-04-26
Fix for disappearing browser when clicking a cookbook's CHANGELOG tab

### LINKED ISSUE #1026: When clicking a cookbook's CHANGELOG tab, browser header disappears
Visit https://supermarket.chef.io/cookbooks/datadog
Click "Changelog"
The Changelog takes over the entire view port, and you can no longer scroll up. Clicking another tab will restore the browser to normal state.

Likely has something to do with CSS - but I'm not clear on that yet.

loganmeetsworld:
Fixes #1026

Note: I was able to replicate part of the problem locally and believe the issue was with the `data-equalizer-watch` foundation class, documented [here](http://foundation.zurb.com/sites/docs/v/5.5.3/components/equalizer.html). I removed the class and the problem went away locally.

### DIFF
--- a/src/supermarket/app/views/cookbooks/_main.html.erb
+++ b/src/supermarket/app/views/cookbooks/_main.html.erb
@@ -1,4 +1,4 @@
-<div class="main" data-equalizer-watch>
+<div class="main">
   <% if cookbook.deprecated? %>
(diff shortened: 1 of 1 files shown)


nellshamrell:
:shipit:

PR #1292, opened 2016-04-27
Include Schedule A/B in body to match hard copy.

### LINKED ISSUE #1288: CCLA Issue
Hello,

I noticed that the CCLA on supermarket mentions a Schedule A and Schedule B but we don’t list these in the CCLA itself. The CCLA is based off the Apache CCLA, and should probably include these at the end:

https://www.apache.org/licenses/cla-corporate.txt
```
Schedule A

   [Initial list of designated employees.  NB: authorization is not
    tied to particular Contributions.]

Schedule B

   [Identification of optional concurrent software grant.  Would be
    left blank or omitted if there is no concurrent software grant.]
```

According to legal at Chef we should include both of these in the Supermarket CCLA.

robbkidd:
Fixes #1288 for new installs.


robbkidd:
@loganmeetsworld Could you please review my seeds change? 😀

loganmeetsworld:
👍  @robbkidd

PR #1293, opened 2016-05-02
Changing supermarket activity stream to report correct user for releases

### LINKED ISSUE #965: Supermarket activity stream reports wrong user for releases
The activity stream for my account reports that v2.5.0 & v2.5.1 of the [zookeeper](https://supermarket.chef.io/cookbooks/zookeeper) cookbook were released by jdinthe503 (aka @jakedavis), yet both of those releases were pushed to Supermarket by myself.

Seems like incorrect attribution, since, while he is the owner of said cookbook, myself and a few others are collaborators, and thus can make new releases of the cookbook.

loganmeetsworld:
Fixes #965 
Updates the schema to include relationship between user and cookbook versions in order to attribute cookbook version uploads correctly, includes tests and view updates to reflect this change.


robbkidd:
Sweeeeeet!

I foresee some trouble for existing `CookbookVersion` records that have a null `user_id`. One option I can think of: provide something like a `CookbookVersion.published_by` method that returns `self.user || cookbook.owner` and use that in retrievals for views

loganmeetsworld:
@robbkidd and I talked today about refactoring this code to apply to any entries for cookbook_uploads made prior to this change. We discussed whether this should be reflected in the database in the `CookbookVersion.user_id` attribute or only on the view with a `CookbookVersion#published_by` method catching any time `CookbookVersion.user_id` is nil. If we wanted to move solely to the database we would make all existing id's in the `CookbookVersion.user_id` column correspond to their parent cookbook's `owner`. We thought that this may create some confusion and could be calamitous under future migrations. After benchmarking the `published_by` method in `CookbookVersion.rb` results showed barely any difference in load time so we decided to stick with this plan. Any other input is welcome, but for now this is where the code stands.

nellshamrell:
👍 pending successful travis build

src/supermarket/app/controllers/api/v1/cookbook_uploads_controller.rb
```diff
@@ -44,7 +44,7 @@ def create
     rescue
       render_not_authorized([t('api.error_messages.unauthorized_upload_error')])
     else
-      cookbook_upload.finish do |errors, cookbook, cookbook_version|
+      cookbook_upload.finish(current_user) do |errors, cookbook, cookbook_version|
```
robbkidd: `#finish` no longer needs `current_user` passed in since `user` gets passed in and set in `CookbookUpload.new`

loganmeetsworld: :+1: fixed that part, but still getting some issues in dev, so looking at the method a little closer.

loganmeetsworld: Ah, found the problem, running tests and pushing changes asap.

PR #1299, opened 2016-05-12
Move the cache_path into the omnibus var dir to avoid warnings

robbkidd:
Previously running reconfigure would yield many WARN-level messages
because the cache was embedded in the cookbook directory.  This moves
the cache to a more appropriate directory and cleans up the old cache
dir.

Sample warnings:
```
[2016-05-12T14:02:32+00:00] WARN: Cookbook 'local-mode-cache' is empty or entirely chefignored at /opt/supermarket/embedded/cookbooks/local-mode-cache
[2016-05-12T14:02:32+00:00] WARN: Cookbook 'local-mode-cache' is empty or entirely chefignored at /opt/supermarket/embedded/cookbooks/local-mode-cache
[2016-05-12T14:02:32+00:00] WARN: Cookbook 'local-mode-cache' is empty or entirely chefignored at /opt/supermarket/embedded/cookbooks/local-mode-cache
[2016-05-12T14:02:32+00:00] WARN: Cookbook 'local-mode-cache' is empty or entirely chefignored at /opt/supermarket/embedded/cookbooks/local-mode-cache
[2016-05-12T14:02:32+00:00] WARN: Cookbook 'local-mode-cache' is empty or entirely chefignored at /opt/supermarket/embedded/cookbooks/local-mode-cache
...
```

Lifted from chef/chef-server#168

TODO:
- ~~Is there a `var_dir` or similar variable available in an omnibus software definition so that the path does not need to be hard coded?~~ No.
- ~~Consider replicating [chef-server's upgrade "migrations"](https://github.com/chef/chef-server/blob/master/omnibus/files/private-chef-upgrades/001/024_cleanup_local_cache_path.rb) to handle cleanup of previous version cruft~~ Not necessary at this time.

### DIFF
--- a/omnibus/config/software/supermarket-cookbooks.rb
+++ b/omnibus/config/software/supermarket-cookbooks.rb
@@ -41,7 +41,7 @@ build do
     open("#{cookbooks_path}/solo.rb", "w") do |file|
       file.write <<-EOH.gsub(/^ {8}/, '')
         cookbook_path   "#{cookbooks_path}"
-        file_cache_path "#{cookbooks_path}/cache"
+        cache_path "/var/opt/supermarket/cache"
         verbose_logging true
         ssl_verify_mode :verify_peer
       EOH
--- a/omnibus/cookbooks/omnibus-supermarket/test/integration/default/serverspec/recipes/default_spec.rb
+++ b/omnibus/cookbooks/omnibus-supermarket/test/integration/default/serverspec/recipes/default_spec.rb
@@ -1,0 +1,11 @@
+require 'spec_helper'
+
+describe 'omnibus-supermarket::default' do
+  describe file(property['supermarket']['var_directory'] + '/cache') do
+    it { should be_directory }
+  end
+
+  describe file(property['supermarket']['install_directory'] + '/embedded/cookbooks/local-mode-cache') do
+    it { should_not exist }
+  end
+end


robbkidd:
Tested a couple upgrades and declaring a different cache_path in the config was sufficient for making the old one disappear. If that is not borne out with further testing of the git-triggered build on staging, I'll add a `directory action :delete` to the omnibus' internal cookbook.

omnibus/config/software/supermarket-cookbooks.rb
```diff
@@ -41,6 +41,7 @@
     open("#{cookbooks_path}/solo.rb", "w") do |file|
       file.write <<-EOH.gsub(/^ {8}/, '')
         cookbook_path   "#{cookbooks_path}"
+        cache_path "/var/opt/supermarket/local-mode-cache"
```
schisamo: I would ensure `file_cache_path` is also set:

```ruby
file_cache_path "/var/opt/supermarket/local-mode-cache"
```

These two config options usually get changed in lock step.

sersut: `node_path` is the other one that might generate some extra files.

robbkidd: @schisamo By lock step, do you mean that both `cache_path` and `file_cache_path` should point to the same directory under `/var/opt/supermarket`? Or should it be something like:

```ruby
file_cache_path "/var/opt/supermarket/cache"
cache_path      "/var/opt/supermarket/cache/local-mode"
```

@sersut All these paths I'm just finding out about!

schisamo: @robbkidd Actually after looking over the `Chef::Config` code it appears just setting `cache_path` is enough as other `*_cache_path` values use this as a base:
https://github.com/chef/chef/blob/master/chef-config/lib/chef-config/config.rb#L263

robbkidd: Looks like a `cache_path` of `/var/opt/supermarket` would work?

* local mode [would make](https://github.com/chef/chef/blob/master/chef-config/lib/chef-config/config.rb#L236) `/var/opt/supermarket/local-mode-cache`
* `file_cache_path` [would make](https://github.com/chef/chef/blob/master/chef-config/lib/chef-config/config.rb#L263) `/var/opt/supermarket/cache`

schisamo: @robbkidd That is how I'm interpreting things also. I guess you can do a test build, install said build and see what happens.

robbkidd:
```
$ sudo ls /var/opt/supermarket
data  etc  lib	nginx  postgresql  rails  redis  ssl
```

Looks like we'd want to make `cache_path '/var/opt/supermarket/cache'` just to make sure something cached doesn't clobber actual data directories.

Or—and this might just be silly talk—`/var/cache/supermarket`?

schisamo: @robbkidd Yeah I would place the cache in its own directory (I think `/var/opt/supermarket/cache` is least surprising). You also want to run a `supermarket-ctl reconfigure` if you want to see a number of the chef-zero stuff get created.

robbkidd:
♻️ 👀 folks. Set `cache_path` and wrote a test to confirm the new place is made and local-mode-cache isn't showing up under the install directory.

schisamo:
:+1:

robbkidd:
@nellshamrell Any thing to add or consider before this merges?

nellshamrell:
:shipit:

PR #1300, opened 2016-05-12
Option to add current owner as a collaborators when transfering ownership

### LINKED ISSUE #1224: Transferring cookbook ownership should make the current owner a contributor
Transferring ownership of a cookbook to a new owner current removes the existing owner entirely.  It should instead make the existing owner a contributor.

loganmeetsworld:
Fixes #1224

Adds checkbox in workflow for transferring ownership that gives option to keep current user on as a collaborator, streamlines the logic action for both admins/collaborators and non-admins/collaborators.

### DIFF
--- a/src/supermarket/app/controllers/transfer_ownership_controller.rb
+++ b/src/supermarket/app/controllers/transfer_ownership_controller.rb
@@ -11,11 +11,13 @@ class TransferOwnershipController < ApplicationController
     @cookbook = Cookbook.with_name(params[:id]).first!
     authorize! @cookbook, :transfer_ownership?
     recipient = User.find(transfer_ownership_params[:user_id])
-    if params[:cookbook][:add_owner_as_collaborator] == "1"
-      add_owner_as_collaborator = true
-    else
-      add_owner_as_collaborator = false
-    end
+
+    add_owner_as_collaborator = if params[:cookbook][:add_owner_as_collaborator] == "1"
+                                  true
+                                else
+                                  false
+                                end
+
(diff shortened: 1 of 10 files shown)


nellshamrell:
Looks good, would love confirmation that it works in a terraform'd Supermarket environment.  Let me know if you'd like help doing this :)

nellshamrell:
To check: whether buttons work in common email clients (i.e. Google Inbox)
My suggestions for this: 1) Check that the buttons that are generated from transferring ownership from supermarket production work in Google inbox 2) Check that buttons that are generated from transferring ownership with this new code work in Google inbox.  If they work in 1), but don't work in 2), then there is a regression change we need to address.  If the do NOT work in 1) or 2)...then that may be something we should look into (could you file a Github issue?) but it's not something that would block this change.

nellshamrell:
@loganmeetsworld - have you had a chance to check out whether the buttons work in Google inbox?  I'm only concerned if the work with the master branch but not this branch.

loganmeetsworld:
@nellshamrell is there a way to prevent emails from going to spam through `postfix` ?

nellshamrell:
Indeed there is.  You will want to set this attribute https://github.com/chef/supermarket/blob/master/omnibus/cookbooks/omnibus-supermarket/attributes/default.rb#L319 to something like your personal email (or any other email address), which will make it less likely to end up in spam.  Let me know if you would like assistance in doing this :)

nellshamrell:
We have tested this together IRL and this is ready to go!

:shipit:

Issue #1294, opened 2016-05-09
Research mounting Fieri as a Rails Engine


nellshamrell:
https://github.com/chef/fieri

robbkidd:
Totally doable. Possibly-reusable-spike in branch https://github.com/chef/supermarket/tree/fieri-as-engine

Fieri has two parts:
- web end point for Supermarket to say "hey, here's a new cookbook version to evaluate"
- a worker to do the evaluation

### Why to make fieri a Rails Engine:

😞  Currently, deploying fieri involves a whole separate cookbook to deploy web services, Sidekiq worker processor, and Redis to a separate host. Comms between Supermarket and separate Fieri are beset—especially in private installs—with problems with FQDN resolution and additional SSL cert needs.

😃  Implementing fieri as a Rails engine that is included ("mounted") within Supermarket removes all the additional deployment and operational effort of a separate codebase.

* Included in Supermarket as a gem, a fieri Rails engine automatically becomes a part of the existing omnibus package.
* No additional deployment needs to provide a load balancer, web server, or SSL cert for a separate web service. The fieri web endpoints become endpoints in already running Supermarket web hosts.
* Automatically uses Supermarket's configured Sidekiq workers and Redis The fieri worker becomes just another job in the Sidekiq job processor already deployed as part of Supermarket.
* Scaling Supermarket web/worker tier also scales Fieri web/workers.

### operational concerns

* disk requirements - Fieri requires temporary disk space to untar and then foodcritic a cookbook tarball, flaws in fieri cleaning up after itself could fill a partition
* running foodcritic on Supermarket web nodes opens web nodes to security vulnerabilities in foodcritic - could be slightly mitigated by deployment/configuration changes that specify which services to run on which nodes. e.g. web nodes have [Sidekiq disabled](https://github.com/chef/supermarket/blob/master/omnibus/cookbooks/omnibus-supermarket/attributes/default.rb#L226) and worker nodes have web services disabled [1](https://github.com/chef/supermarket/blob/master/omnibus/cookbooks/omnibus-supermarket/attributes/default.rb#L99) [2](https://github.com/chef/supermarket/blob/master/omnibus/cookbooks/omnibus-supermarket/attributes/default.rb#L198)

robbkidd:
We have decided to do this for the reasons above. Being worked on #1303.

PR #1303, opened 2016-05-18
Fieri as a Rails engine

robbkidd:
A spike that's become useful following the effort to research engines as an option (#1294)

TODO:
- [x] update omnibus to know about modified/new fieri environment variables
- [x] test inclusion in the omnibus package
- [x] test that routing to `/fieri/{jobs|status}` is disabled if the feature flag is disabled
- [x] default `FIERI_{URL|RESULTS_ENDPOINT}` as `http://localhost:13000/` to have Rails app (mostly talk to itself?


src/fieri/spec/models/cookbook_worker_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      )
+
+      job_id_2 = CookbookWorker.perform_async(
+        'cookbook_artifact_url' => 'http://example.com/apache.tar.gz',
+        'cookbook_name' => 'apache2',
+        'cookbook_version' => '1.2.0'
+      )
+
+      assert Dir.exist?(File.expand_path(Dir.tmpdir, job_id_1))
+      assert Dir.exist?(File.expand_path(Dir.tmpdir, job_id_2))
```
robbkidd: These `expand_path`s ought to be `join`s.

robbkidd: Maybe.

robbkidd:
Reconsidering the current "check for feature flag within fieri itself" business. In the light of a new day, it seems like a lot of code as compared to deciding whether the mount routes to fieri or not. Investigating testing a `ROLLOUT.active? :fieri` constraint on the mount instead so that within the engine, fieri assumes since it is getting requests that it is enabled.

robbkidd:
So far, I've got working the following spec to replace the entirety of [the commit testing the feature flag within the engine itself](https://github.com/chef/supermarket/pull/1303/commits/9892ead499e8148b4a78c28b0b29d7e379cf3a1b).

```ruby
require 'spec_helper'

describe 'fieri routes', type: :request do
  describe 'GET /status' do
    it 'should return a 200' do
      expect(ROLLOUT).to receive(:active?).with(:fieri).and_return(true)
      get fieri.status_path
      expect(response).to have_http_status(200)
    end
    it 'does not exist if fieri feature is disabled' do
      expect(ROLLOUT).to receive(:active?).with(:fieri).and_return(false)
      expect { get fieri.status_path }
        .to raise_error(ActionController::RoutingError)
    end
  end
end
```

Thoughts? Feelings? Not a fantastic test, but also simplifies the code.

src/fieri/.env
```diff
@@ -0,0 +1,4 @@
+FIERI_FOODCRITIC_FAIL_TAGS=any
+FIERI_FOODCRITIC_TAGS=~FC031 ~FC045
+FIERI_RESULTS_ENDPOINT=http://supermarket.getchef.com/api/v1/cookbook-verisons/evaluation
+FIERI_KEY=YOUR_FIERI_KEY
```
loganmeetsworld: Results endpoint needs to be changed to local here, doesn't need to go out to public endpoint, it can just talk to itself. If we ever get to a point where we split up where the web app and workers run, the port will need to change.

src/fieri/.env.test
```diff
@@ -0,0 +1 @@
+REDIS_URL='redis://localhost:6379/2/fieri'
```
robbkidd: Added this to fix weird race-ish conditions when running the tests while also having a development instances of Supermarket and it's Sidekiq workers up. Gives a Fieri test run a kinda dedicated redis database separate from Supermarket's default.

src/supermarket/.env
```diff
@@ -18,7 +18,10 @@ CURRY_SUCCESS_LABEL="Signed CLA"
 PUBSUBHUBBUB_SECRET=YOUR_PUBSUBHUBBUB_SECRET
 NEW_RELIC_APP_NAME=Supermarket
 FEATURES=cla,join_ccla,tools,fieri,announcement,github,collaborator_groups
-FIERI_URL=http://example.com
+FIERI_URL=http://localhost:13000
```
robbkidd: Needs `/fieri/jobs`

src/supermarket/.env
```diff
@@ -18,7 +18,10 @@ CURRY_SUCCESS_LABEL="Signed CLA"
...
 FIERI_KEY=YOUR_FIERI_KEY
+FIERI_FOODCRITIC_FAIL_TAGS=any
+FIERI_FOODCRITIC_TAGS=~FC031 ~FC045
+FIERI_RESULTS_ENDPOINT=http://localhost:13000/api/v1/cookbook-verisons/evaluation
```
robbkidd: VERSIONS!

src/supermarket/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM ruby:2.1.9
```
robbkidd: Get rid of this.

PR #1306, opened 2016-05-20
Do not allow a cookbook version to be destroyed if it is the last version of a cookbook

robbkidd:


src/supermarket/app/controllers/api/v1/cookbook_uploads_controller.rb
```diff
@@ -128,6 +129,12 @@ def destroy_version
 
     begin
       authorize! @cookbook, :destroy?
+      raise VersionMustExist if @cookbook.cookbook_versions.count == 1
+    rescue VersionMustExist
+      error({ error_code: t('api.error_codes.conflict'),
+              error_messages: [t('api.error_messages.only_cookbook_version')],
+              error: t('api.error_messages.only_cookbook_version') },
```
robbkidd: I have not yet found what client might be expecting a collection of `error_messages` (which I copied from other error returns here in the API), but `error` is the message displayed by the `knife` CLI. So added it here to make the error meaningful to an end user attempting to use `knife supermarket` or `knife cookbook site` to delete the last version of a cookbook.

src/supermarket/spec/controllers/api/v1/cookbook_uploads_controller_spec.rb
```diff
@@ -244,6 +244,21 @@
...
+
+      it 'returns a 409 with informative error messages' do
+        versions = CookbookVersion.all.pluck(:version)
+        versions.each do |version|
+          delete :destroy_version, cookbook: cookbook, version: version, format: :json
+        end
+
+        expect(response.status.to_i).to eql(409)
+        expect(JSON.parse(response.body)["error_code"]).to eq('CONFLICT')
+        expect(JSON.parse(response.body)["error_messages"][0]).to eq('You may not DELETE the only version of a cookbook. A cookbook must have at least one version.')
```
robbkidd: I recommend we test for `error` presence in the response body because we _know_ that is the message being used by the knife CLI. 

And what are your thoughts on testing that it matches the translation for the message, rather than the string repeated here? I waffle between making the test declarative but brittle with the exact string or making the test less brittle but maybe more indirect by using `I18n.t('api.error_messages.only_cookbook_version')`.

Yey! Tests!

src/supermarket/app/controllers/api/v1/cookbook_uploads_controller.rb
nellshamrell: 👍

nellshamrell:
This looks beautiful :shipit:

src/supermarket/spec/controllers/api/v1/cookbook_uploads_controller_spec.rb
```diff
@@ -244,6 +244,21 @@
...
+
+      it 'returns a 409 with informative error messages' do
+        versions = CookbookVersion.all.pluck(:version)
+        versions.each do |version|
+          delete :destroy_version, cookbook: cookbook, version: version, format: :json
+        end
+
+        expect(response.status.to_i).to eql(409)
+        expect(JSON.parse(response.body)["error_code"]).to_not eq(nil)
+        expect(I18n.t('api.error_messages.only_cookbook_version')).to eq('You may not DELETE the only version of a cookbook. A cookbook must have at least one version.')
```
robbkidd: I suggest something like
```
expect(JSON.parse(response.body)['error']).to eq(I18n.t('api.error_messages.only_cookbook_version'))
```
 to test that the error displays a labeled error message, but not care about the specific wording of the translation.

robbkidd:
👍

PR #1307, opened 2016-05-23
makes s3 bucket domain name style configurable

nellshamrell:
Fixes #1301 

Before this is deployed, we will need to add this value to the supermarket vault
s3_domain_style: ':s3_path_url'

PR #1322, opened 2016-05-31
reverts changing paperclip to use path urls

### LINKED ISSUE #1301: Investigate error with Berks and recent upgrade
When I upgraded 2 of our prod nodes to 2.6.0, we started seeing these errors when running berks install:

Resolving cookbook dependencies with Berkshelf 4.3.3...
E, [2016-05-16T22:20:46.045779 #3130] ERROR -- : Actor crashed!
OpenSSL::SSL::SSLError: hostname "community-files.opscode.com.s3.amazonaws.com" does not match the server certificate

I believe this is due to this PR: #1281

nellshamrell:
This temporarily fixes #1301

What happened was, when we deployed #1281, we started seeing errors from berks.  This was due to the fact that our prod S3 bucket has "." in the name, which cannot be used with the path style.  #1307 begins to address this, but that PR needs some additional work.  In order to make this shippable for now, I am reversing the s3 url style change.

### DIFF
--- a/src/supermarket/config/initializers/paperclip.rb
+++ b/src/supermarket/config/initializers/paperclip.rb
@@ -37,7 +37,7 @@ end
       )
     else
       options = options.merge(
-        url: ':s3_domain_url'
+        url: ':s3_path_url'
       )
     end
 


robbkidd:
👍  pending Travis tests passing.

PR #1324, opened 2016-06-02
WIP: Add a guard to Supermarket-specific ctl commands to require being run as supermarket service user

### LINKED ISSUE #1314: "requiretty" in sudoers breaks "supermarket-ctl make-admin"
Issue Summary:

If "/etc/sudoers" has "Defaults requiretty", as is the default for RHEL/Centos, the "supermarket-ctl make-admin" command will fail with the "sudo: sorry, you must have a tty to run sudo" error because make-admin.rb has sudo embedded in its code. This happens even while logged in as the root user.

Expected Results:
```
# id
uid=0(root) gid=0(root) groups=0(root)

# supermarket-ctl make-admin demo
demo was not found in Supermarket.  Make sure the user exists in Supermarket before making it an admin.
```

Actual Results:
```
# supermarket-ctl make-admin demo

sudo: sorry, you must have a tty to run sudo
```

Repro Steps (if possible):
```
# echo "Defaults    requiretty" >> /etc/sudoers
# supermarket-ctl make-admin demo

sudo: sorry, you must have a tty to run sudo
```
Request for Developers:

I simply removed the "sudo" command in the following line of code and it fixed the problem. I was able to login as the root user and run the "make-admin" command even when "requiretty" was set because sudo was no longer being used.

[https://github.com/chef/supermarket/blob/master/omnibus/cookbooks/omnibus-supermarket/files/default/ctl-commands/make-admin.rb#L10](https://github.com/chef/supermarket/blob/master/omnibus/cookbooks/omnibus-supermarket/files/default/ctl-commands/make-admin.rb#L10)

```
# grep command_text /opt/supermarket/embedded/service/omnibus-ctl/make-admin.rb
  command_text = "cd /opt/supermarket/embedded/service/supermarket && RAILS_ENV=\"production\" env PATH=/opt/supermarket/embedded/bin bin/rake user:make_admin user=\"#{username}\""
  shell_out = Mixlib::ShellOut.new(command_text)

# id
uid=0(root) gid=0(root) groups=0(root)

# grep requiretty /etc/sudoers
Defaults    requiretty

# supermarket-ctl make-admin demo
demo was not found in Supermarket.  Make sure the user exists in Supermarket before making it an admin.
```

robbkidd:
Fixes #1314

Calling sudo within the ctl command conflicts with a host configuration whose sudoers requires a TTY.
This removes the call to sudo from within the command and displays an error to the user if the ctl
command is being run as an OS user other than the 'supermarket' service user.

### DIFF
--- a/omnibus/cookbooks/omnibus-supermarket/files/default/ctl-commands/console.rb
+++ b/omnibus/cookbooks/omnibus-supermarket/files/default/ctl-commands/console.rb
@@ -1,5 +1,17 @@
+# due to how things are being exec'ed, the CWD will be all wrong,
+# so we want to use the full path when loaded from omnibus-ctl,
+# but we need the local relative path for it to work with rspec
+begin
+  require 'helpers/user_check_helper'
+rescue LoadError
+  require '/opt/supermarket/embedded/service/omnibus-ctl/helpers/user_check_helper'
+end
+
 add_command 'console', 'Enter the rails console for Supermarket', 1 do
-  cmd = "cd /opt/supermarket/embedded/service/supermarket && sudo -u supermarket env PATH=/opt/supermarket/embedded/bin:$PATH bin/rails console production"
+  user_check = UserCheckHelper.new('console')
+  user_check.must_run_as 'supermarket'
+
+  cmd = "cd /opt/supermarket/embedded/service/supermarket && env PATH=/opt/supermarket/embedded/bin:$PATH bin/rails console production"
   exec cmd
   true
 end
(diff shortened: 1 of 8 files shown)


nellshamrell:
Are we requiring them just for our custom ctl commands, or for the built in omnibus ctl commands as well?

robbkidd:
Just the three commands we've added for supermarket. There is no user check for the omnibus-ctl built-in commands which results in permission denied errors on things like `-ctl restart`.

nellshamrell:
Gotcha!  This looks fabulous, then!

robbkidd:
I kind of want to rename the helper. Something like `CtlCommandHelper`.

robbkidd:
The guard does not catch on the `test` and `console` commands, only works for `make-admin` for reasons beyond me at the moment. Marking this PR as WIP until we get that figured out.

nellshamrell:
Merging this one to correct a customer issue, have opened up a new ticket (#1325) to address the test and console commands.

PR #1338, opened 2016-06-14
Fix Quality tab not displaying successful foodcritic feedback.

### LINKED ISSUE #1336: Foodcritic success is hidden :(
In v2.7.1, we are hiding foodcritic success.

On a test node:
```
irb(main):010:0> cb.name
=> "dragons"
irb(main):011:0> cb.cookbook_versions.pluck(:foodcritic_failure)
=> [nil, true, true, true, false]
```

The latest version of the `dragons` cookbook (0.5.0) ​_passes_​ foodcritic tests (`foodcritic_failure` is `false`), but we display "No quality metric results found." in the view. I think the reason for this is in [the view for the content of the Quality tab](https://github.com/chef/supermarket/blob/2.7.1/src/supermarket/app/views/cookbooks/_main.html.erb#L115-L123)  `version.foodcritic_failure.present?` evaluates to `false` when `foodcritic_failure == false`.

```
irb(main):012:0> false.present?
=> false
```

And then the view thinks it does not have any results to display.

### Proposal:

Have fieri post both successful and unsuccessful content for the `foodcritic_feedback` attribute. If there is content, display it. If there is no content (e.g. `version.foodcritic_feedback.nil?`), display the message that there are no results.

Consider deprecating the `foodcritic_failure` attribute.

robbkidd:
Fixes #1336 

foodcritic_failure will be nil if fieri hasn't posted back results. If
fieri has posted results, foodcritic_failure will be true or false. false
means success (argh, double negatives!), but false is also not `present?`
and so the test for presence fails and the view thinks there are no
results.

This flips the test from `present?` to `nil?` and flips the if/else
clauses. This was the smallest change to fix the display of foodcritic
success.

Future TODOs:
- extract the quality section of this partial to its own partial for focused testing with a view spec
- remove the foodcritic_failure column altogether and have fieri post feedback—positive or negative—for this view to display more simply

### DIFF
--- a/src/supermarket/app/views/cookbooks/_main.html.erb
+++ b/src/supermarket/app/views/cookbooks/_main.html.erb
@@ -112,14 +112,14 @@
     <% end %>
     <% if ROLLOUT.active?(:fieri) %>
       <div class="content" id="quality">
-        <% if version.foodcritic_failure.present? %>
+        <% if version.foodcritic_failure.nil? %>
+          <h2>No quality metric results found.</h2>
+        <% else %>
           <% if version.foodcritic_feedback.present? %>
             <pre><%= version.foodcritic_feedback.gsub(/\n/, '<br />').html_safe %></pre>
           <% else %>
             <p><%= version.version %> passed <%= link_to 'Foodcritic', 'http://acrmp.github.io/foodcritic', target: '_blank' %>.
           <% end %>
-        <% else %>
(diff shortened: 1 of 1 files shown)


nellshamrell:
👍

robbkidd:
Considering/doing the TODOs in the very near-term quality metrics work.

Issue #1304, opened 2016-05-19
Default attributes depending on other attributes causing headaches for custom configurations


robbkidd:
The attempts to use node attributes to set other default node attributes within the omnibus internal cookbook is causing some of our configuration headaches.

For example, [we set the default value for `protocol` depending on the value of `force_ssl`](https://github.com/chef/supermarket/blob/master/omnibus/cookbooks/omnibus-supermarket/attributes/default.rb#L329). Because `protocol` and `force_ssl` are evaluated and set here in the defaults, the evaluation for `protocol` never uses an overridden value for `force_ssl`. And so, `ENV['PROTOCOL']` gets set to `https` all the time and then [`Supermarket::Host.full_url` always returns a URL with `https` as the scheme](https://github.com/chef/supermarket/blob/master/src/supermarket/lib/supermarket/host.rb#L4-L8).  

That and `https` being [hardcoded in the nginx site template](https://github.com/chef/supermarket/blob/master/omnibus/cookbooks/omnibus-supermarket/templates/default/rails.nginx.conf.erb#L123) are the two things that trip up a non-SSL install. I was able to get non-SSL Supermarket working in a VM by setting `"protocol": false` directly in `supermarket.json` and manually changing the nginx site config to `http` ​_after_​ a `-ctl reconfigure`.

nellshamrell:
Need to fix defaults - we should never interpolate default values.

nathenharvey:
see also tip 5 here - https://coderanger.net/chef-tips/

PR #1339, opened 2016-06-14
contains calls to outside services - vendors one font, removes another

### LINKED ISSUE #1305: Add ability to disable calls to 3rd parties in view rendering
In airgapped/secure environments, calls out to 3rd parties are not allowed.  As it stands right now, supermarket attempts to reach out to several sites, including googleapis.com and others (https://github.com/chef/supermarket/blob/master/src/supermarket/app/views/layouts/application.html.erb#L17-L22)

Can we get the ability to turn these calls on/off so Supermarket can run error-free inside of secured environments?

nellshamrell:
Fixes #1305 

1) Vendors one external font source
2) Removes the Montserrat font and uses the default font as an accent font (the default fault was already the fallback font for the accent font)
3) This does not alter Google Analytics - as Google Analytics will only work when the GOOGLE_ANALYTICS_ID environmental variable is set.

### DIFF
--- a/src/supermarket/Gemfile
+++ b/src/supermarket/Gemfile
@@ -61,6 +61,7 @@ gem 'sass-rails',   '~> 4.0.4'
 gem 'sass-globbing'
 gem 'compass-rails'
 gem 'uglifier',     '~> 2.7'
+gem 'font-awesome-rails'
 
 group :doc do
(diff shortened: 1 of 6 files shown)


nellshamrell:
Closing, see #1304 for details.

PR #1342, opened 2016-06-20
Quality Metrics CollaboratorsWorker

### LINKED ISSUE #1311: Quality Metrics: An admin can see whether the number of collaborators >= 1 as a quality metric
As an admin user
When I log into Supermarket
And view a cookbook version
And I click on the “Quality” tab
And the number of collaborators is >=1
Then I see that the “Number of collaborators” metric has passed
When the number of collaborators <=1
Then I see that the metric has failed

- [x] Add logic to determine total number of collaborators
- [x] Research interface (api calls?) between Fieri and Supermarket to determine whether this metric has passed
- [x] Expand Supermarket's api to cooperate with Fieri - maybe add additional API client
- [x] Research how to incorporate collaborator results into results record stored for a cookbook version
- [x] Incorporate collaborator results into results record stored for a cookbook version
- [x] When total number of collaborators passes, show happy message
- [x] When total number of collaborators fails, show failure message

loganmeetsworld:
Fixes #1311 

Post-team review to-do list: 
- [x] Need to make the new collaborator worker endpoint into an environment variable if that is the path we are taking
- [x] Update the Fieri `.env` to have the localhost version of the api route (both supermarket and fieri need to have their endpoints checked)
- [x] Internationalize the collaborator feedback string
- [x] Reconsider wording for the current collaborator success and failure message
- [x] Consider requiring collaborator message in supermarket api `cookbook_versions_controller.rb`
- [x] In the cookbook view `main` partial make sure logic is consistent with master re: present vs. nil
- [x] Move from separate `foodcritic` and `collaborator` endpoints in the supermarket api for cookbooks to a new section of the json response for cookbook versions (`/api/v1/cookbooks/dragons/foodcritic` to `/api/v1/cookbooks/dragons/versions/0.1.1`) (Handled by #1363) 
- [x] Reconsider changing the evaluation endpoint, have Fieri append the metric endpoint itself.


src/fieri/spec/dummy/db/schema.rb
```diff
@@ -0,0 +1,15 @@
...
+# Note that this schema.rb definition is the authoritative source for your
+# database schema. If you need to create the application database on another
+# system, you should be using db:schema:load, not running all the migrations
+# from scratch. The latter is a flawed and unsustainable approach (the more migrations
+# you'll amass, the slower it'll run and the greater likelihood for issues).
+#
+# It's strongly recommended that you check this file into your version control system.
+
+ActiveRecord::Schema.define(version: 0) do
+end
```
robbkidd: Was this a carry over from a time in development when the fieri engine was going to be responsible for some quality metric tables?

loganmeetsworld: @robbkidd Looking back, this got committed in a rubocop appease-al for some reason.

robbkidd: The Great and Powerful Rubocop must be Appeased!

src/fieri/.env
```diff
@@ -1,4 +1,5 @@
 FIERI_FOODCRITIC_FAIL_TAGS=any
 FIERI_FOODCRITIC_TAGS=~FC031 ~FC045
-FIERI_RESULTS_ENDPOINT=http://supermarket.getchef.com/api/v1/cookbook-verisons/evaluation
+FIERI_FOODCRITIC_ENDPOINT=http://supermarket.getchef.com/api/v1/cookbook-verisons/evaluation
```
nellshamrell: Should the this be along the lines of cookbook-versions/foodcritic_evaluation, or whatever the new route is?

src/fieri/app/models/collaborator_worker.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def evaluate(cookbook_name)
+    cookbook_json = get_json(cookbook_name)
+    collaborator_count = get_collaborator_count(cookbook_json)
+    sufficient_collaborators?(collaborator_count)
+  end
+
+  def give_feedback(cookbook_name)
+    evaluate(cookbook_name) ? 'This cookbook has sufficient collaborators.' : 'This cookbook does not have sufficient collaborators.'
```
robbkidd: We should consider the wording of the display of the results of this measure. What's the message we'd like to convey about number of collaborators? What would we like to encourage the cookbook author to do? Or the consumer to know?

src/supermarket/.env
```diff
@@ -22,6 +22,6 @@ FIERI_URL=http://localhost:13000/fieri/jobs
 FIERI_KEY=YOUR_FIERI_KEY
 FIERI_FOODCRITIC_FAIL_TAGS=any
 FIERI_FOODCRITIC_TAGS=~FC031 ~FC045
-FIERI_RESULTS_ENDPOINT=http://localhost:13000/api/v1/cookbook-versions/evaluation
+FIERI_FOODCRITIC_ENDPOINT=http://localhost:13000/api/v1/cookbook-versions/evaluation
```
nellshamrell: Is this also a different route now?  food_critic_evaluation?

src/supermarket/app/controllers/api/v1/cookbook_versions_controller.rb
```diff
@@ -76,4 +93,9 @@ def require_evaluation_params
     params.require(:cookbook_version)
     params.require(:foodcritic_failure)
   end
+
+  def require_collaborator_params
+    params.require(:cookbook_name)
+    params.require(:collaborator_failure)
+  end
```
robbkidd: Maybe add `params.require(:collaborator_message)` because in this context, we don't have a case where we don't have a message to display. (Unlike foodcritic, where there may be no findings for the message because it passed.)

src/supermarket/app/views/cookbooks/_main.html.erb
```diff
@@ -112,14 +112,20 @@
     <% end %>
     <% if ROLLOUT.active?(:fieri) %>
       <div class="content" id="quality">
-        <% if version.foodcritic_failure.nil? %>
-          <h2>No quality metric results found.</h2>
-        <% else %>
+        <% if version.foodcritic_failure.present? || version.collaborator_failure.present? %>
```
loganmeetsworld: Make sure that this logic is consistent with the working logic in master re: double negative with the foodcritic failure and three states of a foodcritic response: `nil`, `false`, and `not false`.

src/supermarket/spec/api/cookbook_collaborator_criticism_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    expect(response.status.to_i).to eql(200)
+  end
+
+  it 'returns the collaborator results' do
+    get "/api/v1/cookbooks/#{cookbook.name}/collaborator"
+
+    expect(signature(json_body)).to include(cookbook_collaborator_signature)
+  end
+end
```
loganmeetsworld: Move from having `foodcritic` and `collaborator` endpoints on a cookbook to a quality section within the endpoint for a cookbook version. Information about failures and quality for a given version of a cookbook should be housed in the api endpoint for a cookbook version: For example, within `/api/v1/cookbooks/dragons/versions/0.1.1` there would be a quality section of json.

src/supermarket/spec/controllers/api/v1/cookbook_versions_controller_spec.rb
```diff
@@ -186,4 +186,81 @@
...
+            cookbook_name: cookbook.name,
+            cookbook_version: version.to_param,
+            collaborator_failure: false,
+            collaborator_feedback: 'This cookbook does not have sufficient collaborators.',
+            fieri_key: 'YOUR_FIERI_KEY',
+            format: :json
+          )
+
+          expect(version.reload.collaborator_failure).to eql(false)
+          expect(version.reload.collaborator_feedback).to eql('This cookbook does not have sufficient collaborators.')
```
loganmeetsworld: All of the times we use 'sufficient' language (or whatever language we want) should be in the internationalization section.

src/fieri/.env
```diff
@@ -1,4 +1,5 @@
 FIERI_FOODCRITIC_FAIL_TAGS=any
 FIERI_FOODCRITIC_TAGS=~FC031 ~FC045
-FIERI_RESULTS_ENDPOINT=http://supermarket.getchef.com/api/v1/cookbook-verisons/evaluation
+FIERI_FOODCRITIC_ENDPOINT=http://supermarket.getchef.com/api/v1/cookbook-verisons/foodcritic_evaluation
+FIERI_COLLABORATORS_ENDPOINT=http://localhost:3000/api/v1/cookbook-versions/collaborators_evaluation
```
robbkidd: Upon pondering this some more, I foresee pain in adding a new endpoint ENV var for each metric type. Most of the URL is the same, but would have to be repeated in N settings. As new metrics are added to the system, using them would require installers to add more ENV vars to their configs.

An option:

* Go back to `FIERI_RESULTS_ENDPOINT` and have it default to `http://localhost:3000/api/v1/cookbook-versions`.
* Fieri would append the metric endpoint itself—`foodcritic_evaluation` or `collaborators_evaluation`—to that results endpoint.

This way, an installation of Supermarket with Fieri enabled would only have to provide the URL to their custom domain + `api/v1/cookbook-versions` once.

src/supermarket/config/locales/en.yml
```diff
@@ -96,6 +96,10 @@ en:
       unsubscribe:
         failure: "There was an issue unsubscribing Supermarket from the repo."
         success: "Supermarket is no longer monitoring this repo."
+  quality_metrics:
+    collaborator:
+      success: 'passed with %{collaborators}.'
+      failure: 'Failure: Cookbook has %{collaborators}. A cookbook must have at least %{passing_number} collaborators to pass this metric.'
```
robbkidd: I think these entries ought to go into `src/fieri/config/locales/en.yml` for the engine to own the messages.

Also, I suggest the interpolated variable here be `num_collaborators` for both messages and move the string "collaborators" to display after the number from the CollaboratorWorker calls to I18n to here.

src/fieri/app/models/collaborator_worker.rb
```diff
@@ -22,11 +22,17 @@ def get_collaborator_count(json)
...
   end
 
   def give_feedback(cookbook_name)
-    evaluate(cookbook_name) ? 'This cookbook has sufficient collaborators.' : 'This cookbook does not have sufficient collaborators.'
+    json = get_json(cookbook_name)
+    collaborator_count = get_collaborator_count(json)
+    if evaluate(cookbook_name)
+      I18n.t('quality_metrics.collaborator.failure', :collaborators => collaborator_count.to_s + ' collaborators', :passing_number => '2')
+    else
+      I18n.t('quality_metrics.collaborator.success', :collaborators => collaborator_count.to_s + ' collaborators')
```
robbkidd: Related to other comment: I suggest the interpolated variable here be `num_collaborators` for both messages and move the string "collaborators" to display after the number from here to the I18n strings.

Is `passing_number` something known elsewhere in this class for the evaluation to have occurred?

src/supermarket/config/locales/en.yml
loganmeetsworld: I changed the variable and moved the entries 👍

src/fieri/app/models/collaborator_worker.rb
loganmeetsworld: `passing_number` is just in case we wanted to change that value, and it's only shown/related to a failure message.

loganmeetsworld:
What the failures look like in development:
screen shot 2016-06-30 at 11 50 51 am

loganmeetsworld:
Success UI: 
screen shot 2016-06-30 at 12 53 38 pm

robbkidd:
Purdy.

src/supermarket/config/routes.rb
```diff
@@ -23,10 +23,11 @@
       get 'users/:user' => 'users#show', as: :user
 
       # This was the original route, which has a misspelling (cookbook-verisons rather than cookbook-versions).  Keeping this here so as not to break anyone who is still depending on the original route.
-      post '/cookbook-verisons/evaluation' => 'cookbook_versions#evaluation', constraints: proc { ROLLOUT.active?(:fieri) }
+      post '/cookbook-verisons/foodcritic_evaluation' => 'cookbook_versions#foodcritic_evaluation', constraints: proc { ROLLOUT.active?(:fieri) }
 
       # This route has the correct spelling of cookbook-versions
-      post '/cookbook-versions/evaluation' => 'cookbook_versions#evaluation', as: :cookbook_versions_evaluation, constraints: proc { ROLLOUT.active?(:fieri) }
+      post '/cookbook-versions/foodcritic_evaluation' => 'cookbook_versions#foodcritic_evaluation', as: :cookbook_versions_evaluation, constraints: proc { ROLLOUT.active?(:fieri) }
```
nellshamrell: Should this be /cookbook-versions/evaluation/foodcritic_evaluation  ?

src/fieri/.env
```diff
@@ -1,4 +1,5 @@
 FIERI_FOODCRITIC_FAIL_TAGS=any
 FIERI_FOODCRITIC_TAGS=~FC031 ~FC045
-FIERI_RESULTS_ENDPOINT=http://supermarket.getchef.com/api/v1/cookbook-verisons/evaluation
+FIERI_RESULTS_ENDPOINT=http://supermarket.getchef.com/api/v1/cookbook-versions
```
nellshamrell: Should we make this a different default?  Since this will only work in prod?

src/supermarket/.env
```diff
@@ -18,10 +18,11 @@ CURRY_SUCCESS_LABEL="Signed CLA"
 PUBSUBHUBBUB_SECRET=YOUR_PUBSUBHUBBUB_SECRET
 NEW_RELIC_APP_NAME=Supermarket
 FEATURES=cla,join_ccla,tools,fieri,announcement,github,collaborator_groups
-FIERI_URL=http://localhost:13000/fieri/jobs
+FIERI_URL=http://localhost:3000/fieri/jobs
 FIERI_KEY=YOUR_FIERI_KEY
 FIERI_FOODCRITIC_FAIL_TAGS=any
 FIERI_FOODCRITIC_TAGS=~FC031 ~FC045
-FIERI_RESULTS_ENDPOINT=http://localhost:13000/api/v1/cookbook-versions/evaluation
+FIERI_FOODCRITIC_ENDPOINT=http://localhost:3000/api/v1/cookbook-versions/foodcritic_evaluation
```
nellshamrell: Should these two separate endpoints still be defined here?

nellshamrell: oh, wait, nvm

robbkidd:
👍

nellshamrell:
👍

PR #1343, opened 2016-06-22
Untar all files in a cookbook artifact as binary to avoid encoding issues

robbkidd:
This should fix fieri's CookbookWorker failing with encoding errors when binary files (like other tarballs) are included in a cookbook artifact.


nellshamrell:
:shipit: pending Travis tests passing

loganmeetsworld:
👍

PR #1350, opened 2016-06-24
Air gap feature flag

### LINKED ISSUE #1305: Add ability to disable calls to 3rd parties in view rendering

nellshamrell:
Fixes #1305


nellshamrell:
Confirmed that this works in local dev.

loganmeetsworld:
👍 looks good! Good documentation.

PR #1367, opened 2016-07-01
Refactor fieri env vars

### LINKED ISSUE #1365: Refactor handling of environmental variables in Fieri
We have a few different ways of calling Supermarket endpoints from within Fieri

i.e. ENV['SOME_VARIABLE'] = "https://supermarket_url/cookbooks/"
ENV['SOME_OTHER_VARIABLE'] = "https://supermarket_url/cookbook-versions/"

It would perhaps be better to have something like:
ENV['SUPERMARKET_RAILS_APP'] = "https://supermarket_url"

Then use it like this throughout Fieri:
"#{ENV['SUPERMARKET_RAILS_APP']}/cookbooks/"
"#{ENV['SUPERMARKET_RAILS_APP']}/cookbook-versions/"

nellshamrell:
Fixes #1365


robbkidd:
We should probably include changes to the omnibus internal cookbook defaults, too, to include the new ENV variable and remove the old.

src/supermarket/.env
```diff
@@ -22,7 +22,6 @@ FIERI_URL=http://localhost:3000/fieri/jobs
 FIERI_KEY=YOUR_FIERI_KEY
 FIERI_FOODCRITIC_FAIL_TAGS=any
 FIERI_FOODCRITIC_TAGS=~FC031 ~FC045
-FIERI_RESULTS_ENDPOINT=http://localhost:3000/api/v1/cookbook-versions
-FIERI_COLLABORATORS_REQUEST=http://localhost:3000/api/v1/cookbooks
+FIERI_SUPERMARKET_ENDPOINT=http://localhost:3000
```
robbkidd: After an error from Staging following an upgrade to a nightly version and then some reading on dotenv, I believe that these values in `.env` ought to be defaults appropriate for production (so, in this case, port `13000`). [Development environment overrides](https://github.com/bkeepers/dotenv#multiple-rails-environments) should happen in a dev's `.env.development` or `.env.local`.

nellshamrell: Done in #1368

robbkidd:
👍

PR #1371, opened 2016-07-05
Fix fieri's handling of cookbook tarballs with non-dir/file entries

### LINKED ISSUE #1364: Fieri CookbookArtifact doesn't handle PaxHeader data in tarballs
## The Problem

(From #1359)

`NoMethodError cookbook_artifact.rb in criticize at line 40` shows up when a cookbook is uploaded and returns an error on results when it reads the result feedback is a string. Problem shown with cookbook on this part of the criticism in `cookbook_artifact.rb`: `[result.to_s, result.failed?]`. Error: `NoMethodError: undefined method 'failed?' for String`. Unable to criticize this cookbook (quality tab was thus left blank): https://supermarket.chef.io/cookbooks/alteryx-server#quality

## Background

We need to expand on [the logic used to determine the name of the cookbook under test](https://github.com/chef/supermarket/blob/master/src/fieri/app/models/cookbook_artifact.rb#L77). It attempts to find the first directory level of the entries in the tarball, but assumes that the first entry is an actual file or directory. Some versions of tar, like Windows, put other entries in like PaxHeader to add metadata to the archive (ownership, modified timestamps, etc).

For example:
The cookbook tarball for [alteryx-server v0.11.4] has as its first entry a PaxHeader for the top-level `alteryx-server` directory.

 ```
[29] pry(main)> cba = CookbookArtifact.new('https://s3.amazonaws.com/community-files.opscode.com/cookbook_versions/tarballs/17012/original/alteryx-server.tgz?1467064801','nojobid')
=> #<CookbookArtifact:0x007f86d377ae78
 @archive=#<File:/var/folders/py/gk9z_67s0mjbg0w5m3xl_w180000gn/T/archive20160629-76031-ph2xs0 (closed)>,
 @directory="/var/folders/py/gk9z_67s0mjbg0w5m3xl_w180000gn/T/nojobid/PaxHeader",
 @job_id="nojobid",
 @url="https://s3.amazonaws.com/community-files.opscode.com/cookbook_versions/tarballs/17012/original/alteryx-server.tgz?1467064801",
 @work_dir="/var/folders/py/gk9z_67s0mjbg0w5m3xl_w180000gn/T/nojobid">

[30] pry(main)> tar_extract = Gem::Package::TarReader.new(Zlib::GzipReader.open(cba.archive.path))
=> #<Gem::Package::TarReader:0x007f86d2741660 @init_pos=0, @io=#<Zlib::GzipReader:0x007f86d2741930>>

[31] pry(main)> tar_extract.rewind
=> 0

[32] pry(main)> tar_extract.each do |entry|
[32] pry(main)*   puts entry.full_name
[32] pry(main)*   puts entry.directory?
[32] pry(main)*   puts entry.file?
[32] pry(main)*   # puts entry.read
[32] pry(main)* end

PaxHeader/alteryx-server
false
false
alteryx-server/
true
false
alteryx-server/PaxHeader/.gitignore
false
false
alteryx-server/.gitignore
false
true
...
```

### 1 Make looking up the folder of the "first entry" smarter, use a directory or file, not a PaxHeader

When [looking up the "root" of the tarball](https://github.com/chef/supermarket/blob/master/src/fieri/app/models/cookbook_artifact.rb#L77), iterate over entries until we find one of type `directory?` or `file?`. Then do the rewind as we already do to perform the extraction of files.

### 2 Use mixlib-archive to extract the tarball

Chef has already solved the PaxHeader problem in [mixlib-archive](https://github.com/chef/mixlib-archive). We could opt to not care about the first directory of the tarball and run foodcritic from the Worker's `work_dir` instead of making a subdirectory within the `work_dir` based on the name of the cookbook as determined by the first directory within the tarball.

robbkidd:
Use mixlib-archive to extract cookbook tarball in Fieri

* fixes extracting tarballs containing non-dir/file entries like PaxHeader
  metadata
* extraction logic is in mixlib-archive, so delete our similar-but-insufficient
  implementation of untarring in Ruby
* moves downloading and untarring into a prep method so we're not doing
  things that can fail during object initialization
* removes the need for a sub-directory within the work_dir (not sure we ever
  had that need) and therefore the need to lookup the first entry in the tarball
  to figure out what to name that sub-directory; just use the work_dir for
  critiquing
* removes the existence checks from the foodcritic_worker_spec because once
  the worker completes, it cleans up and removes the directory; jobs get unique
  directories to work in because they get dirs named with the job id

Closes #1364


thommay:
You _might_ want to go to mixlib-archive 0.2.0 and disallow the tarball from setting permissions ( https://github.com/chef/chef/pull/5080/files#diff-d2d6bb7266330a4a3813a6e01996895dR338 ) otherwise this looks good.

robbkidd:
@thommay To confirm, the `perms: false, ignore: /^\.$/` bit in `#extract()`?

thommay:
da!

robbkidd:
OK. Tested quite a bit locally.

:eyes: @nellshamrell @thommay

thommay:
:+1: LGTM

PR #1381, opened 2016-07-21
Refactor transfer process

### LINKED ISSUE #1380: Make current owner a collaborator when transferring doesn't work
There seems to be some sort of failure condition when transferring cookbooks with the "make current owner a collaborator" checkbox enabled. I've transferred a few that had just the one owner without issue, but when I transfer a cookbook with a large number of owners and the new owner is already a collaborator it seems to fail. I'm not sure exactly what the condition is without creating some fake cookbooks, but under certain conditions it seems to fail.

loganmeetsworld:
Fixes #1380

### DIFF
--- a/src/supermarket/app/models/cookbook.rb
+++ b/src/supermarket/app/models/cookbook.rb
@@ -156,13 +156,12 @@ class Cookbook < ActiveRecord::Base
 
   def transfer_ownership(initiator, recipient, add_owner_as_collaborator = false)
     original_owner = owner
+
     if initiator.is?(:admin) || collaborator_users.include?(recipient)
       update_attribute(:user_id, recipient.id)
 
       if add_owner_as_collaborator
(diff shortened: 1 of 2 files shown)


src/supermarket/app/models/cookbook.rb
```diff
@@ -155,10 +155,13 @@ def sorted_cookbook_versions
   #
 
   def transfer_ownership(initiator, recipient, add_owner_as_collaborator = false)
+    original_owner = owner
+
     if initiator.is?(:admin) || collaborator_users.include?(recipient)
       update_attribute(:user_id, recipient.id)
 
       if add_owner_as_collaborator
+        create_new_collaborator(original_owner) unless initiator == original_owner
```
nellshamrell: Just run this for the original_owner, not initiator.

robbkidd:
👍

PR #1382, opened 2016-07-21
Adding collaborator triggers fieri

### LINKED ISSUE #1378: Adding a new collaborator needs to triger quality re-eval
Collaborators isn't version specific. If someone gets a bad score and they go about fixing it by adding additional collaborators they shouldn't have to upload a new cookbook. We should re-eval the bad score when they add more collaborators.

loganmeetsworld:
Fixes #1378 pending testing it out in a Terraform environment.

### DIFF
--- a/src/supermarket/app/controllers/collaborators_controller.rb
+++ b/src/supermarket/app/controllers/collaborators_controller.rb
@@ -41,11 +41,7 @@ class CollaboratorsController < ApplicationController
       add_group_members_as_collaborators(resource, collaborator_params[:group_ids]) if collaborator_params[:group_ids].present?
 
       if collaborator_params[:resourceable_type] == 'Cookbook'
-        if ROLLOUT.active?(:fieri) && ENV['FIERI_URL'].present?
-          FieriNotifyWorker.perform_async(
-            resource.latest_cookbook_version.id
-          )
-        end
+        perform_fieri(resource)
       end
 
       redirect_to resource, notice: t('collaborator.added')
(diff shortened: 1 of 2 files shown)


src/supermarket/app/controllers/collaborators_controller.rb
```diff
@@ -40,6 +40,10 @@ def create
 
       add_group_members_as_collaborators(resource, collaborator_params[:group_ids]) if collaborator_params[:group_ids].present?
 
+      if collaborator_params[:resourceable_type] == 'Cookbook'
```
nellshamrell: Change to testing whether class of resource is a Cookbook?

loganmeetsworld:
Tested in Terraform and it works.

nellshamrell:
👍 Merge at will!

src/supermarket/app/controllers/collaborators_controller.rb
loganmeetsworld: @nellshamrell @robbkidd can I get a 👍  on not changing this or are there opinions that I should change it to `resource.class == 'Cookbook'`? Definitely willing to do either way.

nellshamrell: Give it a 20 min spike and see if changing it breaks any tests?

loganmeetsworld: It worked and reads a little better so we'll go with this.

PR #1386, opened 2016-07-27
Add a HOME to db migrate command in omnibus

robbkidd:
Our calls to `bundle exec` in the omnibus install need to have a HOME set because of an updated [rb-readline that requires an `ENV['HOME']` to be present](https://github.com/ConnorAtherton/rb-readline/issues/8). This is related to 28344e92c243974c991093474d90b64a7fe37b72, where a HOME was added to the runit startup scripts for rails and sidekiq. This is the only other `bundle exec` in the omnibus cookbook and so it needs a HOME, too.

Presumably fixes chef-cookbooks/supermarket-omnibus-cookbook#41, but we're currently lacking an automated reproduction of the problem and test of the fix.

### DIFF
--- a/omnibus/cookbooks/omnibus-supermarket/recipes/database.rb
+++ b/omnibus/cookbooks/omnibus-supermarket/recipes/database.rb
@@ -80,6 +80,9 @@ end
 execute 'database schema' do
   command 'bundle exec rake db:migrate db:seed'
   cwd node['supermarket']['app_directory']
-  env 'RAILS_ENV' => 'production'
+  environment({
+    'RAILS_ENV' => 'production',
+    'HOME' => node['supermarket']['app_directory']
+  })
   user node['supermarket']['user']
 end


robbkidd:
@nellshamrell @irvingpop :eyes: Thoughts on how to test when we have not been able to reproduce the problem?

irvingpop:
👍 on the fix

Regarding the repro, it's going to be tough to build an automated test given that we've attempted to simply unset `HOME` and that doesn't work.

As a manual repro case, we could try building a node that connects to hosted Chef and:
1. Initially just runs the `chef-client` cookbook to spin up a daemonized chef client
2. modify the run-list of the node to add `supermarket-omnibus-cookbook` to the runlist.
3. Watch the next CCR to see what happens

robbkidd:
👌 I can try that sadly-not-automated test tomorrow.

robbkidd:
@jorhett hot-patched this change on the host giving him trouble and the chef run succeeded.

So, I'm as 👍  as I think I can be.

robbkidd:
With two thumbs, a successful if sadly-not-automated test, the automated tests not failing, and the collective brains agreeing that "just set a HOME" is the solution, I hereby merge this fix.

PR #1387, opened 2016-07-28
Refactor env var host fqdn

### LINKED ISSUE #1232: host attribute doesn't pick up custom value from fqdn attribute
When a user wants to use a CNAME for their Supermarket server then they correctly configure `node.set['supermarket_omnibus']['config']['fqdn']` in the wrapper cookbook. However, when `supermarket-ctl reconfigure` gets run the [host attribute](https://github.com/chef/omnibus-supermarket/blob/2.3.3/cookbooks/omnibus-supermarket/attributes/default.rb#L324) is not getting set to the new value from the custom `fqdn` attribute.

```
$ jq '.supermarket.fqdn, .supermarket.host' /etc/supermarket/supermarket-running.json
"supermarket-test.lxc"
"8383-supermarket.lxc"
```

This causes problems especially when logging in because the Supermarket redirect url that gets passed to OC-ID is the original FQDN of the Supermarket server instead of the custom CNAME.

You can see in the screenshot below that the `redirect_uri` points to `8383-supermarket.lxc` when it should point to `supermarket-test.lxc`.

screen shot 2016-02-24 at 4 36 49 pm

This can be worked around by setting `node.set['supermarket_omnibus']['config']['host']` to the CNAME FQDN and running `supermarket-ctl reconfigure`.

Please make sure that the [host attribute](https://github.com/chef/omnibus-supermarket/blob/2.3.3/cookbooks/omnibus-supermarket/attributes/default.rb#L324) properly picks up a custom value for the `fqdn` attribute.

nellshamrell:
Fixes #1232


src/supermarket/spec/models/universe_spec.rb
```diff
@@ -4,15 +4,32 @@
...
+      expect(ENV['HOST']).to be_nil
+      ENV['FQDN'] = "localhost"
+
+      opts = {
+        port: 6060,
+        protocol: 'https'
+      }
+
+      expect(Universe.protocol_host_port(opts)).to eql("https://#{ENV['FQDN']}:6060")
+      expect(Universe.download_url(cookbook, version, "https://#{ENV['FQDN']}:6060")).to eql("https://#{ENV['FQDN']}:6060/api/v1/cookbooks/redis/versions/1.3.1/download")
```
robbkidd: How would you feel about using a wild non-defaulty string for the FQDN in this test? Like:

```
expected_fqdn = 'this_is_my_supermarket_there_are_many_like_it_but_this_one_is_mine'
ENV['FQDN'] = expected_fqdn

...

expect(Universe.protocol_host_port(opts)).to eql("https://#{expected_fqdn}:6060")
expect(Universe.download_url(cookbook, version, "https://#{expected_fqdn}:6060")).to eql("https://#{expected_fqdn}:6060/api/v1/cookbooks/redis/versions/1.3.1/download")
```

Juuuuuust in case `localhost` sneaks into tests or something as a default.

nellshamrell: That was my original thought.  Unfortunately, it was causing some of the existing VCR fixtures to fail...though I suppose I can replace those as well.

src/supermarket/spec/models/curry/repository_spec.rb
```diff
@@ -22,4 +22,22 @@
     expect(repository.owner).to eql('chef')
     expect(repository.name).to eql('supermarket')
   end
+
+  describe '#subscribe!' do
+    let(:expected_fqdn) { 'bunniesareadorable' }
```
robbkidd: ✨ 🐰 ✨

src/supermarket/spec/models/curry/repository_spec.rb
```diff
@@ -22,4 +22,22 @@
...
+
+    before do
+      ENV['FQDN'] = expected_fqdn
+    end
```
robbkidd: I think this may need a variable to capture current FQDN and a paired `after` block (or an `around`)  to set it back. Else we have only moved the sneaky global to another place.

nellshamrell: It seems to set it back, since it is effectively in a before each block, rather than a before all block (before all is where we would need to unset it).  It *seems* like it resets ENV['FQDN'] back to localhost, but I will confirm that.

nellshamrell:
Tried this on a Terraform'd Supermarket environment and logins seem to work fine :)

robbkidd:
One day we'll clean up all the direct setting of ENV vars in the specs. But maybe not today.

👍

PR #1391, opened 2016-07-29
Switch to using with_name scope to handle downcasing incoming cookbook names

robbkidd:
Use the [`Cookbook.with_name`](https://github.com/chef/supermarket/blob/master/src/supermarket/app/models/cookbook.rb#L20-L24) scope to handle downcasing the cookbook name to find. Otherwise, the chained methods fail for cookbooks whose names contain capitals
when fieri posts results.

```
NoMethodError: undefined method `cookbook_versions' for nil:NilClass
```


robbkidd:
👀  @nellshamrell ?

nellshamrell:
🚀 🍰

PR #1394, opened 2016-08-02
Add a helper that returns all platforms know to supermarket

tas50:
Creates a single place for us to add new platforms. This will be used by the "any" concept in a bit

### DIFF
--- a/src/supermarket/app/helpers/supported_platforms_helper.rb
+++ b/src/supermarket/app/helpers/supported_platforms_helper.rb
@@ -11,25 +11,44 @@ module SupportedPlatformsHelper
   #
   def known_platforms
     %(aix
+      alpine
       amazon
+      arch
       cento
+      cloudlinux
       debian
       fedora
       freebsd
       gentoo
+      hpux
+      ibm_powerkvm
+      ios_xr
+      linuxmint
       mac_os_x
+      netbsd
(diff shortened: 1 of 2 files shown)


nellshamrell:
👍

tas50:
I should mention that I have no local supermarket to test this on (and I'd love to get that setup sometime)

src/supermarket/app/helpers/supported_platforms_helper.rb
```diff
@@ -1,4 +1,39 @@
...
+      oracle
+      redhat
+      ubuntu
+      scientific
+      smartos
+      solaris
+      suse
+      windows
+      zlinux)
+  end
```
robbkidd: I have a not-fully-formed feeling that this canonical list of known platform names should live deeper, like as a class constant in the `SupportedPlatforms` model. That could still be bubbled up to views through this helper method.

robbkidd: I _do_ agree that we should have the canonical list _somewhere_ to use consistently.

tas50:
After realizing we were missing a lot of platforms this is going to take some rework. See 5a58c1ea23012e3a7f54632c7b2174e7fafb8cdc for the large list

PR #1400, opened 2016-08-03
adds code to send foodcritic version and tags to supermarket and to d…

### LINKED ISSUE #1354: Quality Metrics: Research whether we can show what rules passed
In Foodcritic on the Quality tab

nellshamrell:
…isplay them in supermarket

Fixes #1354

### DIFF
--- a/src/fieri/app/models/foodcritic_worker.rb
+++ b/src/fieri/app/models/foodcritic_worker.rb
@@ -35,7 +35,7 @@ class FoodcriticWorker
   private
 
   def foodcritic_info
-    "Run with Foodcritic Version #{FoodCritic::VERSION} with tags #{ENV['FIERI_FOODCRITIC_TAGS']}"
+    "Run with Foodcritic Version #{FoodCritic::VERSION} with tags #{ENV['FIERI_FOODCRITIC_TAGS']} and failure tags #{ENV['FIERI_FOODCRITIC_FAIL_TAGS']}"
   end
 
   def format_feedback(feedback, status)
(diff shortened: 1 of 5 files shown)


nellshamrell:
@robbkidd if you have a moment when you are at Surge Con, would love if you could give this a quick review.  This is blocking #1401

src/fieri/app/models/foodcritic_worker.rb
```diff
@@ -23,12 +23,26 @@ def make_post(params, feedback, status)
...
     unless response.is_a? Net::HTTPSuccess
       error_msg = "Unable to POST Foodcritic Evaluation of #{params[cookbook_name]} " + response.message
       raise error_msg
     end
   end
+
+  private
+
+  def foodcritic_info
+    "Run with Foodcritic Version #{FoodCritic::VERSION} with tags #{ENV['FIERI_FOODCRITIC_TAGS']}"
```
robbkidd: We ought to include `ENV['FIERI_FOODCRITIC_FAIL_TAGS']` too because that's the list of categories run against (along with the `FIERI_FOODCRITIC_TAGS` which default to some we turn off).

robbkidd: At least according to my understanding of the foodcritic help and [how we call foodcritic in fieri](https://github.com/chef/supermarket/blob/master/src/fieri/app/models/cookbook_artifact.rb#L43-L45) with both `-f` and `-t`:
```
foodcritic [cookbook_paths]
    -t, --tags TAGS                  Check against (or exclude ~) rules with the specified tags.
    -f, --epic-fail TAGS             Fail the build based on tags. Use 'any' to fail on all warnings.
```

nellshamrell: Fixed!

nellshamrell:
@robbkidd integrated the change you suggested around including the failure tags used with foodcritic

nellshamrell:
Closing and re-opening another PR using the branch associated repo, rather than my fork.

PR #1405, opened 2016-08-06
Use the new Foodcritic URL

tas50:
Avoid a redirect and help the right URL get to the top in Google search results


robbkidd:
:+1:

PR #1406, opened 2016-08-06
Consistent naming

tas50:
We're mixing up plural vs. singular in the wording


robbkidd:
:+1:

PR #1409, opened 2016-08-12
Upgrade from Rails 4.1 to 4.2

robbkidd:
There are a couple remaining issues not covered currently by our tests:

- [x] Can't migrate from a new database. We've got [an index created by a migration from 2014](https://github.com/chef/supermarket/blob/2032d961a31f5b21cca6e4b2177b5a4f2f6be097/src/supermarket/db/migrate/20140206202928_create_curry_unknown_pull_request_committer_table.rb#L5) that generates a name for the index that is too long. ActiveRecord apparently only started erroring on that in 4.2. `db:schema:load` works which is why Travis is passing.
- [x] Rails 4.2 doesn't like [what we're doing to create cookbooks with cookbook versions](https://github.com/chef/supermarket/blob/2032d961a31f5b21cca6e4b2177b5a4f2f6be097/src/supermarket/db/seeds.rb#L252-L264) in db/seeds.rb. Because the cookbook isn't saved yet, the cookbook_version's `first_or_create` fails.

PR #1410, opened 2016-08-18
Allow omnibus test-kitchen tests to run in parallel

robbkidd:
Previously, the exact file path to a package file had to be set
for a version. Sadly, for RHEL, that file path includes the platform
version (e.g. el6 vs el7). This updates the kitchen config and test
recipe to only require a version number for the package to be tested
and computes the file path.

package_path's default is based on the kitchen config's
synced_folders.

Note, you'll still need to have a build for a particular platform
stowed in omnibus/pkg prior to running kitchen for the package to
be available within the TK machine.

Signed-off-by: Robb Kidd <rkidd@chef.io>

### DIFF
--- a/omnibus/cookbooks/omnibus-supermarket/.kitchen.yml
+++ b/omnibus/cookbooks/omnibus-supermarket/.kitchen.yml
@@ -2,7 +2,7 @@
 driver:
   name: vagrant
   network:
-    - ['private_network', { ip: '192.168.33.33' }]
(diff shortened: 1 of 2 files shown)


robbkidd:
Awaiting the Rails upgrade in #1409 to pass the bundle-audit.

robbkidd:
#1409 merged. Rebased and force-pushed this branch. Testing.

Issue #1335, opened 2016-06-13
Quality Metrics: An Admin can see a metric that a normal user cannot


nellshamrell:
As an admin user
When I log into Supermarket
And view a cookbook version
And I click the "Quality" tab
Then I can see a hidden metric

As a non-admin user
When I log into Supermarket
And view a cookbook version
And I click the "Quality" tab
Then I cannot see the hidden metric
- [ ] Research how we use rollout and whether we can use it to show or not show a metric  
- [ ] Implement way to hide metrics from regular users, but show to admin users

robbkidd:
Possibly implement in conjunction with "published to Supermarket" quality metric as a low-hanging fruit metric to only show to admins.

nellshamrell:
I did some research on this, and I think we have approached the point where we need to refactor from storing metric results in the cookbook version to using a separate metrics table (so we can more easily mark certain metrics as public and some as admin only).

## What the system does now

Currently, we store both the foodcritic and collaborator method results in the cookbook_version table:

![img_1054](https://cloud.githubusercontent.com/assets/813007/18494733/fd6d8f9a-79cc-11e6-8d7e-ab5334bba2df.JPG)

And this is what the current workflow looks like for running metrics after a user uploads a cookbook

![img_1055](https://cloud.githubusercontent.com/assets/813007/18494749/1f2c76d2-79cd-11e6-9473-a8499acda012.JPG)

## What a refactor would look like

I scoped out a refactor for this, and it is fortunately not as complicated as I thought it was.

I propose we create two new tables - one for quality metrics, and one for metrics results.  They would look like this:

![img_1056](https://cloud.githubusercontent.com/assets/813007/18494764/411a9f76-79cd-11e6-9817-caf1fcb61c5b.JPG)

And then the only part of the workflow that changed would be storing the metric results in Supermarket's DB, how we run them would not have to change.

![img_1057](https://cloud.githubusercontent.com/assets/813007/18494777/56838daa-79cd-11e6-90fd-da42f2cc30f7.JPG)

This would necessitate a migration to move the current metric data out of the cookbook version records and into the metrics results records, but that is doable and should be straightforward to test with a snapshot of our DB.

@robbkidd, what do you think?

nellshamrell:
And I don't know why it turned the images sideways, but if you click on them it will open a new tab with them in the correct orientation.

nathenharvey:
👍

nellshamrell:
[...]

PR #1411, opened 2016-08-18
standardize navmenu hide/show behavior; upgrade Paperclip

### LINKED ISSUE #1587: Paperclip version shipped in the RPM packages is out of date
The RPMs available on https://downloads.chef.io/supermarket/2.9.30 have an old version of Paperclip ruby gem. This is causing S3_URLS_EXPIRE to be set as an absolute time value instead of a time period in seconds for the link to expire. Paperclip version >= 5 sets the correct method arguments (expires_in) instead of (expires) see http://www.rubydoc.info/github/marcel/aws-s3/AWS%2FS3%2FS3Object.url_for for more details

robbkidd:
The menu links in the top nav and in the left-side drawer should hide and appear at the same size. Previously, there was a range of browser widths ("small" to "medium") that would make the top nav disappear before the drawer nav would appear.

Also includes `capybara-screenshot` to produce screenshots when capybara tests fail. This gem was helpful in immediately showing what the problem was with CLA signing tests failing in my dev environment:

screen shot 2016-08-13 at 1 18 29 pm

The usermenu was getting bumped down to _under_ the div below the navmenu, in this case the alert box. Capybara would complain about not being able to click on usermenu items because the alert div was in the way. And the usermenu was getting bumped down because the browser width in my test environment was in a range where the top navmenu didn't hide, causing the usermenu to get bumped down.

Update: also fixes #1587 

## TODO

+ [x] ensure that _all_ of the necessary Rails app config files—e.g. [`.env.production`](https://github.com/chef/supermarket/blob/41663566ec1a19ff775d1cbab8f754a4feca87cb/omnibus/cookbooks/omnibus-supermarket/recipes/app.rb#L26-L37)—are written out prior to [the omnibus migrating the database](https://github.com/chef/supermarket/blob/41663566ec1a19ff775d1cbab8f754a4feca87cb/omnibus/cookbooks/omnibus-supermarket/recipes/database.rb#L80-L88) or move the database migration into the app recipe
+ [x] squash the commits on #1307 and cherry-pick the result into this PR so that all the Paperclip+S3 changes are delivered together
+ ~add a validation to audit_config for `s3_domain_style` being either `:s3_domain_url` or `:s3_path_url`~
+ [x] add a validation to audit_config to error if `:s3_domain_style` is `:s3_domain_url` and `:s3_bucket` contains a period `.`


robbkidd:
Yea. This is going to take a bit more work.

src/supermarket/config/initializers/paperclip.rb
```diff
@@ -21,7 +21,8 @@
       },
       path: path,
       bucket: ENV['S3_BUCKET'],
-      s3_protocol: ENV['PROTOCOL']
+      s3_protocol: ENV['PROTOCOL'],
+      s3_region: ENV['S3_REGION']
```
robbkidd: `s3_region` must be provided for the Paperclip config after the upgrade of the underlying `aws-sdk`. So, an `s3_region` is now required as part of the Supermarket config for storing cookbooks in an S3 bucket.

I suspect we might need to communicate this via a major version bump. As of the writing of this comment on the PR, a Supermarket instance that is currently using an S3 bucket whose administrators do not add an `s3_region` to their settings will _stop_ using the bucket and start storing to local filesystem after an upgrade to a version with this code.

Possibly there could be more checks added around the presence of `S3_BUCKET`, `S3_ACCESS_KEY_ID`, `S3_SECRET_ACCESS_KEY`, and `S3_REGION`. Maybe if any exist, but not all of them, we throw an error at prevent startup?

robbkidd: @nathenharvey @nellshamrell Do you have opinions on this? 

Major version bump to communicate "hey, you gotta do something or behavior might change on you"?

Let it ride with a minor or patch and call the change out on the release notes and announcement?

Do more checking of S3 settings being complete?

nellshamrell:
Did some troubleshooting of the broken specs.  

Looks like they broke with [this commit](https://github.com/chef/supermarket/pull/1411/commits/e1d5ab8d950aadaba76895192b21fc1b2d7ad6be)

And, after further troubleshooting, looks like they broke with changing 

app/assets/stylesheets/appheader.scss
```
@media #{$small-down} {
```

to 

app/assets/stylesheets/appheader.scss
```
@media #{$medium-down} {
```

Is there another way to get the same effect without changing that style specifically?

robbkidd:
### When signed out

hide the mainnav - logged out

### When signed in

hide the mainnav - logged in

nellshamrell:
Nice!

robbkidd:
I reckon what's left is how to communicate to Supermarket+S3 users that they _must_ supply an `s3_region` to the config or any new cookbooks will get stored on disk.

robbkidd:
Proposal for dealing with the new required-if-you're-S3ing `s3_region`:

* major version bump
* add logic to the Supermarket's paperclip initializer that will abort startup if 1-3, but not all 4 `s3_*` config settings are set to non-blank.

## Impact

### private supermarkets using local storage

> None. 

Cookbook artifacts continue to be stored and served from local disk.

### private supermarkets using S3

> Supermarket will fail to start if not given required S3 configs.

Must provide an `S3_REGION` config attribute along with the other attributes—`S3_BUCKET`, `S3_ACCESS_KEY_ID`, and `S3_SECRET_ACCESS_KEY`—that are already set in their environments. If a private supermarket upgrades to the version including this change without an `S3_REGION` set, the web (`rails`) and background jobs (`sidekiq`) services will fail to start and emit a `FATAL` log message about the missing config attribute(s).

@nathenharvey Can we get your weigh in on this approach?

nathenharvey:
Let's push ahead for a major version bump.  That version should include

* If the S3 environment is not fully configured, the Supermarket should fail to start and should provide proper guidance on how to fix that.
* The work being done for #1335 as that will likely involve a migration

Let's revisit during queue replenishment on 26-Sep-2016 to check on status of #1335 as we don't want this branch hanging around for too long.

src/supermarket/config/initializers/paperclip.rb
robbkidd: Confirmed with @nathenharvey that the S3 settings must be absent or complete else the service should fail to start and provide useful feedback

src/supermarket/spec/lib/supermarket/s3_config_audit_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'spec_helper'
+
+describe Supermarket::S3ConfigAudit do
+  context '#use_s3?' do
+    it 'is true if required S3 configs are given in environment' do
+      mock_env = {
+        'S3_BUCKET' => 'bettergetabucket',
+        'S3_ACCESS_KEY_ID' => 'thisismyidtherearemanylikeitbutthisoneismine',
```
nellshamrell: this_is_my_human

src/supermarket/config/initializers/paperclip.rb
```diff
@@ -21,7 +17,8 @@
...
     }
 
     if ENV['S3_PRIVATE_OBJECTS'].present?
```
robbkidd: Tangentially, I've noticed these `.present?` checks which make me sad. I submit that they should be some sort of truthy check on the value instead of just having the variable present in the environment.

The following is probably going to confuse a human at some point.
```
> export S3_PRIVATE_OBJECTS="false"

> be rails c
Loading development environment (Rails 4.2.7.1)
[1] pry(main)> ENV['S3_PRIVATE_OBJECTS']
=> "false"
[2] pry(main)> ENV['S3_PRIVATE_OBJECTS'].present?
=> true
```

nellshamrell: that is a good point

nellshamrell: we could do something like !ENV['S3_PRIVATE_OBJECTS'].nil?

robbkidd: That's essentially the same as current behavior.

```
> export S3_PRIVATE_OBJECTS="false"

> be rails c
Loading development environment (Rails 4.2.7.1)
[1] pry(main)> ENV['S3_PRIVATE_OBJECTS']
=> "false"
[2] pry(main)> !ENV['S3_PRIVATE_OBJECTS'].nil?
=> true

Comment (robbkidd) [inline src/supermarket/config/initializers/paperclip.rb]: Surveying the current uses of `.present?`

* [`S3_PATH`](https://github.com/chef/supermarket/blob/66189112c5b95689866df565a6bda2679291dea9/src/supermarket/config/initializers/paperclip.rb#L10) :: 🆗  because if a path is given, we want to use it.
* [`S3_PRIVATE_OBJECTS `](https://github.com/chef/supermarket/blob/66189112c5b95689866df565a6bda2679291dea9/src/supermarket/config/initializers/paperclip.rb#L27) :: ⚠️  because it will cause unexpected behavior if someone sets it to `"false"`.
* [`CDN_URL `](https://github.com/chef/supermarket/blob/66189112c5b95689866df565a6bda2679291dea9/src/supermarket/config/initializers/paperclip.rb#L33) :: 🆗  because if the URL is given, we want to use it.

Boiled down, I'm of the opinion that `.present?` is good for environment variables that have values to use to build paths, authenticate to things and such. But `S3_PRIVATE_OBJECTS` is acting like a feature flag and it doesn't take a value of "false" in a way that would turn that feature off.

**Comment by robbkidd** [comment]
Followup on what `NOPE, not gonna start like that` will look like at the moment.

@silverbadge Do you have any thoughts on what more could be shown to you as a private Supermarket administrator that could make your life better in this situation?

@nathenharvey Thoughts? I picture including something along the lines of what's below in the release notes when this eventually goes out.

### What the End User Sees When Visiting the Site

A big `502 Bad Gateway` web page and nothing else. 

The web service `rails` to which nginx is reverse proxying is not responding to requests at this point.

### Services

The `rails` and `sidekiq` services will show stopped or running but only for a very short time.

```
$ sudo supermarket-ctl status
run: nginx: (pid 2699) 652s; run: log: (pid 2687) 655s
run: rails: (pid 4882) 2s; run: log: (pid 2706) 652s
run: sidekiq: (pid 4871) 5s; run: log: (pid 2710) 652s
```

### Logs

`/var/log/supermarket/rails/current` and `/var/log/supermarket/sidekiq/current` will both have repeated errors as the web and background worker services attempt to start, fail, and restart.

```
[timestamp] Supermarket::S3ConfigAudit::IncompleteConfig: Got some, but not all, of the required S3 configs. Must provide none or a
ll of ["S3_BUCKET", "S3_ACCESS_KEY_ID", "S3_SECRET_ACCESS_KEY", "S3_REGION"] to configure cookbook storage in an S3 bucket.
```

**Comment by robbkidd** [comment]
For instance, something more that could maybe be done is to detect the incomplete S3 configuration within the omnibus package's `reconfigure` step which would fail Chef converge instead of letting Chef succeed and then failing service startup.

**Comment by nathenharvey** [comment]
A failing Chef Run when you don't have the proper values would be best.  The Supermarket would continue to run and the the chef-client run would provide a meaningful error message.

**Comment by robbkidd** [comment]
I'll start working on a config check within the omnibus reconfigure.

can do

**Comment by silverbadge** [comment]
The supermarket app log are pinpointing the root cause for the web server failure! Spot on 🥇 
It is reasonable to fail if the config is incorrect or incomplete. Thanks for fixing the S3 region issue as well as the S3 expiring url by upgrading paperclip in this PR.

**Comment by robbkidd** [comment]
Here is a sample of what I've got in development for a Chef run's `supermarket-ctl reconfigure` failing with a similar error right shebang at the beginning. No state changes made by the omnibus reconfigure with an invalid S3 config.

Note: we use lowercase string keys for these config values when setting them on the node object.

```
---- Begin output of supermarket-ctl reconfigure ----
STDOUT: [2017-04-05T18:10:07+00:00] WARN: Port 8889 not available
[2017-04-05T18:10:07+00:00] INFO: Started chef-zero at chefzero://localhost:8890 with repository at /opt/supermarket/embedded
  One version per cookbook

[2017-04-05T18:10:07+00:00] INFO: Forking chef instance to converge...
[2017-04-05T18:10:08+00:00] INFO: *** Chef 12.13.37 ***
[2017-04-05T18:10:08+00:00] INFO: Platform: x86_64-linux
[2017-04-05T18:10:08+00:00] INFO: Chef-client pid: 13998
[2017-04-05T18:10:10+00:00] INFO: HTTP Request Returned 404 Not Found: Object not found: chefzero://localhost:8890/nodes/default-ubuntu-1404
[2017-04-05T18:10:10+00:00] INFO: Setting the run_list to ["recipe[omnibus-supermarket::default]"] from CLI options
[2017-04-05T18:10:10+00:00] INFO: Run List is [recipe[omnibus-supermarket::default]]
[2017-04-05T18:10:10+00:00] INFO: Run List expands to [omnibus-supermarket::default]
[2017-04-05T18:10:10+00:00] INFO: Starting Chef Run for default-ubuntu-1404
[2017-04-05T18:10:10+00:00] INFO: Running start handlers
[2017-04-05T18:10:10+00:00] INFO: Start handlers complete.
[2017-04-05T18:10:10+00:00] INFO: HTTP Request Returned 404 Not Found: Object not found:
[2017-04-05T18:10:12+00:00] INFO: Loading cookbooks [omnibus-supermarket@1.1.0, enterprise@0.10.3, omnibus@5.1.5, runit@3.0.5, sysctl@0.8.1, unicorn@2.1.2, openssl@7.0.1, build-essential@8.0.0, chef-sugar@3.4.0, chef-ingredient@2.0.1, git@6.0.0, homebrew@3.0.0>
[2017-04-05T18:10:13+00:00] INFO: Storing updated cookbooks/omnibus-supermarket/recipes/config.rb in the cache.

================================================================================
Recipe Compile Error in /var/opt/supermarket/cache/cache/cookbooks/omnibus-supermarket/recipes/default.rb
================================================================================

Supermarket::Config::IncompleteConfig
-------------------------------------
Got some, but not all, of the required S3 configs. Must provide none or all of ["s3_bucket", "s3_access_key_id", "s3_secret_access_key", "s3_region"] to configure cookbook storage in an S3 bucket.

Cookbook Trace:
---------------
  /var/opt/supermarket/cache/cache/cookbooks/omnibus-supermarket/libraries/config.rb:66:in `audit_config'
  /var/opt/supermarket/cache/cache/cookbooks/omnibus-supermarket/recipes/config.rb:40:in `from_file'
  /var/opt/supermarket/cache/cache/cookbooks/compat_resource/files/lib/chef_compat/monkeypatches/chef/run_context.rb:347:in `load_recipe'
  /var/opt/supermarket/cache/cache/cookbooks/compat_resource/files/lib/chef_compat/monkeypatches/chef/run_context.rb:303:in `block in include_recipe'
  /var/opt/supermarket/cache/cache/cookbooks/compat_resource/files/lib/chef_compat/monkeypatches/chef/run_context.rb:302:in `each'
  /var/opt/supermarket/cache/cache/cookbooks/compat_resource/files/lib/chef_compat/monkeypatches/chef/run_context.rb:302:in `include_recipe'
  /var/opt/supermarket/cache/cache/cookbooks/omnibus-supermarket/recipes/default.rb:20:in `from_file'
  /var/opt/supermarket/cache/cache/cookbooks/compat_resource/files/lib/chef_compat/monkeypatches/chef/run_context.rb:347:in `load_recipe'
```

nellshamrell:
This looks good to me!

PR #1414, opened 2016-08-29
Replace Librarian tab with Policyfile

tas50:
We should make it easier for users to cut and paste if using Policyfile

### DIFF
--- a/src/supermarket/app/models/user.rb
+++ b/src/supermarket/app/models/user.rb
@@ -2,7 +2,7 @@ class User < ActiveRecord::Base
   include Authorizable
   include PgSearch
 
-  ALLOWED_INSTALL_PREFERENCES = %w(berkshelf knife librarian)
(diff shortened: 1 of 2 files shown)


nathenharvey:
👍

robbkidd:
@nellshamrell and I discussed this and we have concluded that it would be a better incremental change to _add_ Policyfile as an install preference and installation tab. I'll work on that today.

robbkidd:
What a combined Berkshelf/Librarian/Policyfile tab could look like.

screen shot 2016-09-07 at 9 24 30 am

Eh ... Looks weird. But it's also a little weird to show the same install content across three separate tabs. Opinions?

robbkidd:
With Policyfile as a separate tab to show declaring a source for the cookbook.

screen shot 2016-09-07 at 12 49 26 pm

tas50:
👍

PR #1426, opened 2016-09-11
Bump berkshelf gem version

tas50:


robbkidd:
👍  but this and #1427 will probably need reconciliation with both changing `Gemfile.lock`. My preference is for #1427 to merge first.

PR #1427, opened 2016-09-12
Upgrade PostgreSQL to 9.3.14

robbkidd:
[9.3.7](https://www.postgresql.org/docs/current/static/release-9-3-7.html):

* CVE-2015-3165: Avoid possible crash when client disconnects just
  before the authentication timeout expires
* CVE-2015-3166: Improve detection of system-call failures and
  information disclosure with out-of-memory situations and buffer
  overflows.
* CVE-2015-3167: uniformly report decryption failures as "Wrong key or
  corrupt data" to avoid the risk of aiding attackers in recovering keys
  from other systems

[9.3.10](https://www.postgresql.org/docs/current/static/release-9-3-10.html):

* CVE-2015-5289: Guard against stack overflows in json parsing
* CVE-2015-5288: Fix contrib/pgcrypto to detect and report too-short
  crypt() salts

[9.3.11](https://www.postgresql.org/docs/current/static/release-9-3-11.html):

* CVE-2016-0773: Very large character ranges in bracket expressions
  could cause infinite loops in some cases, and memory overwrites in
  other cases.
* CVE-2007-4772: A more complete fix for an old fix to regex compiler
  handling loops.
* CVE-2016-0766: Mitigate a PL/Java bug.

[9.3.14](https://www.postgresql.org/docs/current/static/release-9-3-14.html):

* CVE-2016-5423: possible mis-evaluation of nested CASE-WHEN expressions
* CVE-2016-5424: Fix client programs' handling of special characters in
  database and role names. ... considered security fixes because crafted
  object names containing special characters could have been used to
  execute commands with superuser privileges the next time a superuser
  executes pg_dumpall or other routine maintenance operations.

### DIFF
--- a/omnibus/Gemfile.lock
+++ b/omnibus/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/chef/omnibus-software.git
-  revision: 09cf045c2e7622b05348855d7495470777f95df2
+  revision: 8eeb0bb45b01b833389c662932c9fab9042f328b
   specs:
     omnibus-software (4.0.0)
       chef-sugar (>= 3.4.0)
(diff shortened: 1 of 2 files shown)


robbkidd:
👍 

Tested a couple permutations of upgrading Supermarket 2.3.8 and 2.3.15 to a build of this branch.

PR #1433, opened 2016-09-15
Refactor quality metrics

nellshamrell:
This lays the groundwork for #1335 - it refactors our code around quality metrics to 1) Make them easier to add and maintain and 2) More easily allow us to define some metrics as viewable by all users of Supermarket and some by only Supermarket admins


nellshamrell:
This is now ready for review.

src/supermarket/db/migrate/20160914215547_move_quality_metric_results.rb
```diff
@@ -0,0 +1,26 @@
+class MoveQualityMetricResults < ActiveRecord::Migration
+  def change
+    foodcritic_qm = QualityMetric.create!(name: 'Foodcritic')
+    collab_num_qm = QualityMetric.create!(name: 'Collaborator Number')
```
robbkidd: I wonder if we need something to create these in `seeds.rb`, too, for systems built up with `db:schema:load` (I can think of the test environments, [in travis](https://github.com/chef/supermarket/blob/be7bf10424daa999bee8e5dfac4e87d14b786abe/src/supermarket/lib/tasks/travis.rake#L7) and [in automate](https://github.com/chef/supermarket/blob/be7bf10424daa999bee8e5dfac4e87d14b786abe/.delivery/build_cookbook/recipes/unit.rb#L56)).

nellshamrell: That is a good point - adding!

nellshamrell: Done!

nellshamrell:
Building package with this code, then building test infra and testing the migrations on it.

nellshamrell:
Blocked by #1440

robbkidd:
I'm 👍  on this pending our tests and timing of applying this to a test clone of production data.

Also, I recall we're considering releasing this and #1411 together as a major version bump to help communicate the "you should do stuff to make sure you keep working" nature of both.

nellshamrell:
Confirmed that this works on staging infra!

src/supermarket/app/controllers/api/v1/cookbook_versions_controller.rb
```diff
@@ -112,4 +117,13 @@ def require_collaborator_params
...
   end
+
+  def create_metric(cookbook_version, quality_metric, failure, feedback)
+    MetricResult.create!(
+      cookbook_version: cookbook_version,
+      quality_metric: quality_metric,
+      failure: failure,
+      feedback: feedback
+    )
+  end
```
robbkidd: What are the advantages to wrapping the call to `MetricResult.create!` in a controller method over calling it directly in the controller action?

src/supermarket/app/controllers/cookbook_versions_controller.rb
```diff
@@ -25,6 +25,7 @@ def show
     @collaborators = @cookbook.collaborators
     @supported_platforms = @version.supported_platforms
     @owner_collaborator = Collaborator.new resourceable: @cookbook, user: @owner
+    @metric_results = @version.metric_results
```
robbkidd: Did you happen to play around with `.includes(:metric_results)` in queries? We might start considering that for this and `support_platforms` in this controller and in `CookbooksController` to collect what we need in one pass to the database.

src/supermarket/app/views/api/v1/cookbook_versions/_cookbook_version.json.jbuilder
```diff
@@ -4,14 +4,16 @@ json.version cookbook_version.version
...
-    json.feedback @cookbook.collaborator_feedback
+if cookbook_version.metric_results.any?
+  json.quality do
+    json.foodcritic do
+      json.failed foodcritic_metric_result(cookbook_version).failure
+      json.feedback foodcritic_metric_result(cookbook_version).feedback
+    end
+    json.collaborator do
+      json.failed collaborator_num_metric_result(cookbook_version).failure
+      json.feedback collaborator_num_metric_result(cookbook_version).feedback
```
robbkidd: I think each of the `<type>_metric_result()` calls to the helper in this JSON view triggers a call to the database.  The `@metric_results` instance variable is populated with all the cookbook_version's metrics. Maybe we could test for `@metric_results.any?` and iterate over them?

src/supermarket/db/migrate/20160914215547_move_quality_metric_results.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    CookbookVersion.all.each do |cookbook_version|
```
robbkidd: At the risk of triggering the need to test the migration of production data again, consider:

```diff
- CookbookVersion.all.each do |cookbook_version|
+ CookbookVersion.where("foodcritic_failure IS NOT NULL or collaborator_failure IS NOT NULL").each do |cookbook_version|
```

The difference in production is substantial:

```
> CookbookVersion.count
=> 16855
> CookbookVersion.where("foodcritic_failure IS NOT NULL or collaborator_failure IS NOT NULL").count
=> 9447
```

src/supermarket/db/seeds.rb
```diff
@@ -336,4 +338,12 @@
...
     callback_url: ENV['PUBSUBHUBBUB_CALLBACK_URL'] || 'https://seeded.example.com/curry/pr-updates'
   ).first_or_create!
+
+  unless(QualityMetric.where(name: 'Foodcritic').any?)
+    QualityMetric.create!(name: 'Foodcritic')
+  end
+
+  unless(QualityMetric.where(name: 'Collaborator Number').any?)
+    QualityMetric.create!(name: 'Collaborator Number')
+  end
```
robbkidd: 🤔  I suspect these "ensure QualityMetrics exists" in seeds should probably be towards the top and outside of the check for development environment. That would follow the philosophy of making seeds responsible for ensuring the creation of The Standard Set of Things that should exist in any running instance of Supermarket.

src/supermarket/app/controllers/api/v1/cookbook_versions_controller.rb
nellshamrell: To avoid some duplication within the #foodcritic_evaluation and #collaborators_evaluation methods.  I'm fine either doing this in each controller or doing it within a sub method like this, seemed a little dryer.

src/supermarket/app/controllers/cookbook_versions_controller.rb
nellshamrell: Good point, giving it a try

robbkidd: Oof. Yea. `.includes()` would need to move to the query on the model, somewhere in the onion layers of `Cookbook`.

I'd like to rephrase this review comment as "maybe we can refine this later." 😄

src/supermarket/app/controllers/api/v1/cookbook_versions_controller.rb
robbkidd: OK. By my reading, the duplication prevented is not having to type the keyword arguments which I suppose if we suspect they might change in future, this way we've got fewer places to update them.

src/supermarket/app/views/api/v1/cookbook_versions/_cookbook_version.json.jbuilder
robbkidd: The [change](https://github.com/chef/supermarket/commit/ea91d0d80462306f7526735ee231bf4cd5413c9a) ...

```diff
-if cookbook_version.metric_results.any?
+if @cookbook_version_metrics.any?
```

... helps. But  I think the call to `find_by` in [foodcritic_metric_result](https://github.com/chef/supermarket/pull/1433/files#diff-44a50391e9501944d79581f30b09ddcdR76) and [collaborator_num_metric_result](https://github.com/chef/supermarket/pull/1433/files#diff-44a50391e9501944d79581f30b09ddcdR88) will do more look-up queries.

src/supermarket/spec/factories/quality_metric.rb
```diff
@@ -0,0 +1,4 @@
+FactoryGirl.define do
+  factory :quality_metric, class: 'QualityMetric' do
+  end
+end
```
nellshamrell: It was indeed.  Fixed and done!

src/supermarket/app/views/api/v1/cookbook_versions/_cookbook_version.json.jbuilder
nellshamrell: I don’t know that there is a way to fully avoid a trip to the database.
I could do something along the lines of:
```ruby
 json.quality do
      json.foodcritic do
       json.failed @cookbook_version_metrics.where(quality_metric: QualityMetric.foodcritic_metric).first.failure
       json.failed @cookbook_version_metrics.where(quality_metric: QualityMetric.foodcritic_metric).first.feedback
     end
     json.collaborator do
       json.failed @cookbook_version_metrics.where(quality_metric: QualityMetric.collaborator_num_metric).first.failure
       json.failed @cookbook_version_metrics.where(quality_metric: QualityMetric.collaborator_num_metric).first.feedback
     end
```

robbkidd: 🤔  Yea. I don't know immediately how to avoid them, either. Tinkering ...

src/supermarket/app/views/api/v1/cookbook_versions/_cookbook_version.json.jbuilder
```diff
@@ -4,14 +4,16 @@ json.version cookbook_version.version
...
 json.file api_v1_cookbook_version_download_url(cookbook_version.cookbook, cookbook_version)
-json.quality do
-  json.foodcritic do
-    json.failed @cookbook.foodcritic_failure
-    json.feedback @cookbook.foodcritic_feedback
-  end
-  json.collaborator do
-    json.failed @cookbook.collaborator_failure
-    json.feedback @cookbook.collaborator_feedback
+if !@cookbook_version_metrics.nil? && @cookbook_version_metrics.any?
```
robbkidd: We can consider using `.present?` in a future refactor (if we don't opt to iterate).

robbkidd:
👍  Let's push this latest state over to the pipeline for the testin's and the mergin's.

PR #1452, opened 2016-10-11
fixes bug where downloading a cookbook version would break if the coo…

nellshamrell:
…kbook version had one metric but not the other associated with it


src/supermarket/app/views/api/v1/cookbook_versions/_cookbook_version.json.jbuilder
```diff
@@ -4,18 +4,15 @@ json.version cookbook_version.version
...
-  json.quality do
-    json.foodcritic do
-      json.failed foodcritic_metric_result(cookbook_version).failure
-      json.feedback foodcritic_metric_result(cookbook_version).feedback
-    end
-    json.collaborator do
-      json.failed collaborator_num_metric_result(cookbook_version).failure
-      json.feedback collaborator_num_metric_result(cookbook_version).feedback
-    end
+  json.quality_metrics @cookbook_version_metrics do |metric|
```
robbkidd: I think the iteration done by the builder thingie used here alone will take care of the `!nil?` and `any?` case. If `@cookbook_version_metrics` is `nil` or `[]`, the iteration will be a no-op.

Whether that is correct or not, I suggest the `if` above also include (or only be) a feature flag check for fieri being enabled.

nellshamrell: Agreed, will add that and test.

nellshamrell: Done!

src/supermarket/spec/api/cookbook_version_spec.rb
```diff
@@ -99,10 +101,39 @@
...
+
+      context 'when the fieri feature is not active' do
+        before do
+          allow(ROLLOUT).to receive(:active?).with(:fieri).and_return(false)
+        end
 
-        expect(signature(json_body)).to include(quality_metrics)
+        it 'does not return quality metrics for the cookbook version' do
+          get "/api/v1/cookbooks/#{cookbook.name}/versions/#{cookbook_version.version}"
+          expect(JSON.parse(response.body)).to_not include(quality_metrics)
```
nellshamrell: We could do that...but I think that this is a little clearer and I don't think it costs us much in terms of time and memory.

src/supermarket/spec/api/cookbook_version_spec.rb
```diff
@@ -99,10 +101,39 @@
...
+
+        it 'returns quality metrics for the cookbook version' do
+          get "/api/v1/cookbooks/#{cookbook.name}/versions/#{cookbook_version.version}"
+          expect(JSON.parse(response.body)).to include(quality_metrics)
+        end
+
+        context 'when the cookbook has only the foodcritic metric' do
+          before do
+            collab_result.destroy
+            expect(cookbook_version.metric_results.count).to eq(1)
```
nellshamrell: That was meant for development only, removing.

src/supermarket/spec/api/cookbook_version_spec.rb
robbkidd: OK.

src/supermarket/spec/api/cookbook_version_spec.rb
nellshamrell: Fixed!

robbkidd:
👍  Pending test success.

PR #1472, opened 2016-11-18
fix seeds not knowing about a column we just added in a migration

### LINKED ISSUE #1471: QualityMetrics admin_only should only allow true or false
By default, a QM should be admin_only = false, not admin_only = nil

robbkidd:
I'm still confused about why when running `db:migrate db:seeds` together in a single rake command that the `QualityMetric` creation in seeds didn't know about the new column created by the migration. But `reset_column_information` forces a refresh from the schema in the database and gets us past the error.

As a bonus ...

Fixes #1471 by adding a default value of `false` to the `admin_only` column on `QualityMetric` and disallows NULL values. @nellshamrell and I agreed it was safe to modify this migration because it had not yet been included in release of Supermarket, so not production or staging systems would have seen and executed the migration's previous state.

Bonus bonus: learned about [ActiveRecord's `merge()`](http://api.rubyonrails.org/classes/ActiveRecord/SpawnMethods.html#method-i-merge) and applied it judiciously to our open/admin-only scopes.

### DIFF
--- a/src/supermarket/app/models/metric_result.rb
+++ b/src/supermarket/app/models/metric_result.rb
@@ -2,7 +2,7 @@ class MetricResult < ActiveRecord::Base
   belongs_to :cookbook_version
   belongs_to :quality_metric
 
-  scope :open, -> { joins(:quality_metric).where('quality_metrics.admin_only IS NULL') }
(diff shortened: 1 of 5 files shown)


nellshamrell:
👍

PR #1477, opened 2016-11-29
Add License Quality Metric

### LINKED ISSUE #1473: Implement "Cookbook has an open source license" metric
https://github.com/chef-cookbooks/cookbook-quality-metrics/blob/master/quality-metrics/qm-003-license.md

nellshamrell:
Fixes #1473


nellshamrell:
This is now ready for review

src/fieri/app/controllers/fieri/jobs_controller.rb
```diff
@@ -20,5 +22,10 @@ def job_params
       end
       params
     end
+
+    def cookbook_version_response
+      uri = URI.parse("#{ENV['FIERI_SUPERMARKET_ENDPOINT']}/api/v1/cookbooks/#{params['cookbook_name']}/versions/#{params['cookbook_version']}")
+      Net::HTTP.get(uri)
+    end
```
robbkidd: I would expect this lookup to occur within the LicenseWorker job being scheduled by hitting this endpoint, not by the endpoint itself prior to job scheduling.

src/fieri/app/models/license_worker.rb
```diff
@@ -0,0 +1,43 @@
+require 'sidekiq'
+require 'net/http'
+
+class LicenseWorker
+  include ::Sidekiq::Worker
+
+  # Acceptable licenses were determined and documented in
+  # https://github.com/chef-cookbooks/cookbook-quality-metrics/blob/master/quality-metrics/qm-003-license.md
+  ACCEPTABLE_LICENSES = ['Apache 2.0', 'MIT', 'GNU Public License 2.0', 'GNU Public License 3.0'].freeze
```
robbkidd: I suspect we're going to see a number of variations on these strings. I wonder if we should consider updating QM003 to use [the SPDX list of license identifiers](https://spdx.org/licenses/) like Habitat's metadata does. 

@nathenharvey Do you have any opinions about which way to encourage cookbook authors to go for consistent license string formatting?

src/fieri/app/models/license_worker.rb
```diff
@@ -0,0 +1,43 @@
...
+      cookbook_version: JSON.parse(version_json)['version'],
+      license_failure: license_failure?(version_json),
+      license_feedback: license_feedback(cookbook_name, license_failure?(version_json))
+    )
+  end
+
+  private
+
+  def license_failure?(version_json)
+    license = JSON.parse(version_json)['license']
```
robbkidd: `version_json` is getting parsed multiple times. Maybe a local variable e.g. `version_info` within `#perform()` could store the parsed JSON and then `version_info['version']` and `version_info['license']` could be passed to the worker's utility methods?

src/fieri/app/models/license_worker.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    ACCEPTABLE_LICENSES.include?(license) ? false : true
```
robbkidd: Is a ternary here more clear than ...?

```ruby
!ACCEPTABLE_LICENSES.include?(license)
```

_squints_ Maybe. That `!` really does not jump out.

src/fieri/app/models/license_worker.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def license_feedback(cookbook_name, failure)
+    failure == true ? "#{cookbook_name} needs a valid open source license\n#{acceptable_licenses_string}" : ''
+  end
+
+  def acceptable_licenses_string
```
robbkidd: I think this could be made a little bit simpler.

```ruby
def acceptable_licenses_string
  "Acceptable licenses include #{ACCEPTABLE_LICENSES.join(', ')}."
end
```

```
[1] pry(main)> ACCEPTABLE_LICENSES = ['Apache 2.0', 'MIT', 'GNU Public License 2.0', 'GNU Public License 3.0']
=> ["Apache 2.0", "MIT", "GNU Public License 2.0", "GNU Public License 3.0"]
[2] pry(main)> "Acceptable licenses include #{ACCEPTABLE_LICENSES.join(', ')}."
=> "Acceptable licenses include Apache 2.0, MIT, GNU Public License 2.0, GNU Public License 3.0."
```

If the building of the string is reduced to that, it could even get moved back into `#license_feedback()`. And maybe a positive version of the feedback string could be provided:

```ruby
def license_feedback(cookbook_name, failure)
  if failure
    "#{cookbook_name} does not have a valid open source license.\nAcceptable licenses include #{ACCEPTABLE_LICENSES.join(', ')}."
  else
    "#{cookbook_name} has a recognized open source license."
  end
end
```

src/supermarket/app/controllers/api/v1/cookbook_versions_controller.rb
```diff
@@ -1,7 +1,7 @@
 class Api::V1::CookbookVersionsController < Api::V1Controller
   before_action :check_cookbook_name_present, only: [:foodcritic_evaluation, :collaborators_evaluation, :publish_evaluation]
-  before_action :check_authorization, only: [:foodcritic_evaluation, :collaborators_evaluation, :publish_evaluation]
-  before_action :find_cookbook_version, only: [:foodcritic_evaluation, :collaborators_evaluation, :publish_evaluation]
+  before_action :check_authorization, only: [:foodcritic_evaluation, :collaborators_evaluation, :publish_evaluation, :license_evaluation]
+  before_action :find_cookbook_version, only: [:foodcritic_evaluation, :collaborators_evaluation, :publish_evaluation, :license_evaluation]
```
robbkidd: This controller is starting to look ripe for an overhaul. All of the `_evaluation` endpoints are about creating `MetricResult` resources on a `CookbookVersion`. Starting to smell something like an `Api::V1::MetricResultsController`. I'd really like to consider a refactor ...or renovation... of this controller in this go-around.

src/fieri/app/controllers/fieri/jobs_controller.rb
nellshamrell: I thought through that...and I think we have more QM coming down the line which will need the cookbook version json.  I'd rather request it once here, then pass it to the worker.  In the future, I would like to refactor the metrics which operation on the cookbook json to request it in the controller and pass it to the worker.

src/fieri/app/models/license_worker.rb
nellshamrell: I am open to that.

src/fieri/app/models/license_worker.rb
nellshamrell: Done!

src/fieri/app/models/license_worker.rb
nellshamrell: Done!

src/supermarket/app/controllers/api/v1/cookbook_versions_controller.rb
robbkidd: Agree on renovating in a separate PR next.

src/fieri/app/models/license_worker.rb
robbkidd: @nellshamrell and I discussed and are OK with moving ahead with this list while we have the metric flipped to admin-only. But we should decide on a canonical list of license identifiers to use before making this metric public.

src/fieri/app/controllers/fieri/jobs_controller.rb
robbkidd: Gotcha. We'll figure that out during the renovation, mebbe.

robbkidd:
A selective rebase'n'squash would clean up this set of changes, too.

src/fieri/app/models/license_worker.rb
robbkidd: Help text from both `knife cookbook create` and `chef generate cookbook`:

```
-I LICENSE, --license LICENSE
Sets the license. Valid values are all_rights, apache2, mit, gplv2, or gplv3. Default value: all_rights.
```

License strings that appear in metadata from [`knife cookbook create`](https://github.com/chef/chef/blob/master/lib/chef/knife/cookbook_create.rb#L419-L430) match the list specified by QM003 and are as we have implemented the license check:

```
"Apache 2.0", "GNU Public License 2.0", "GNU Public License 3.0", "MIT"
```

License strings that appear in metadata from `chef generate cookbook`:

```
apache2 mit gplv2 gplv3
```

So I suggest at a minimum we add these short versions (as they appear in the help text and in `chef generate`'d metadata). Whether we move to using SPDX short identifiers is probably a bigger question that spans a number of ecosystem tools.

PR #1481, opened 2016-12-01
Include short versions of the open source license strings

robbkidd:
These are the strings that are accepted as license options in `chef generate cookbook -I <license>`. These short versions are also the strings that appear in the metadata of `chef generate`'d cookbooks. (See discussion on [1477](https://github.com/chef/supermarket/pull/1477#discussion_r90479895) about licenses in the generators.)

TODO/CONSIDER:
- [x] ~explicit fixtures and tests for these short version strings?~
- [x] ~any adjustments to the failure feedback in displaying a now-longer list?~

### DIFF
--- a/src/fieri/app/models/license_worker.rb
+++ b/src/fieri/app/models/license_worker.rb
@@ -6,7 +6,10 @@ class LicenseWorker
 
   # Acceptable licenses were determined and documented in
   # https://github.com/chef-cookbooks/cookbook-quality-metrics/blob/master/quality-metrics/qm-003-license.md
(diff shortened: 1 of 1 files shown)


nellshamrell:
👍 

I think we are ok without adding additional tests for the new strings, as we are mainly testing that if a license is not in the ACCEPTABLE_LICENSES constant then the test fails.  I think we have that covered with several test cases.

I would go ahead and not adjust the failure feedback just yet, let's release it as an admin only metric and get feedback from the supermarket admins.

Issue #1484, opened 2016-12-01
Remove JIRA name from Supermarket profile


tas50:
We currently include the users JIRA name on their Supermarket profile. It seems like we're probably safe to remove this at this point.

See tim.smith here:
https://supermarket.chef.io/users/tas50

PR #1486, opened 2016-12-02
include labels for profile form fields

### LINKED ISSUE #1485: Field in profile management page are confusing
The current design of the fields in the manage profile page makes it so you can't see what the fields are for unless you remove the content from the field. There's no name for each field. It makes sense when you initially create your profile since the fields are empty, but if you go to update something it's rather confusing.

screen shot 2016-12-01 at 2 55 53 pm

robbkidd:
Labels will help a human know what these fields are when they come back
to this form months after adding values that replaced the placeholder
text.

Fixes #1485

### New User With Empty Fields

screen shot 2016-12-02 at 10 24 00 am

### User Who Entered Some Things

screen shot 2016-12-02 at 2 26 22 pm

### DIFF
--- a/src/supermarket/app/views/profile/edit.html.erb
+++ b/src/supermarket/app/views/profile/edit.html.erb
@@ -24,12 +24,15 @@
 
           <fieldset>
             <div class="company-field">
+              <%= f.label :company %>
               <%= f.text_field :company, placeholder: 'Company', title: 'Company' %>
             </div>
             <div class="twitterusername-field">
+              <%= f.label :twitter_username, "Twitter Username" %>
               <%= f.text_field :twitter_username, placeholder: 'Twitter Username', title: 'twitter username' %>
             </div>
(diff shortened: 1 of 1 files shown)


robbkidd:
@tas50 Update the description on this PR with screenshots of empty fields and filled-in fields.

robbkidd:
Also bonus'd in fix for #1484 to remove asking for or displaying a user's JIRA username since I was already tinkering with the Manage Profile form.

robbkidd:
In the ordering of things, I'm planning to rebase this against master once #1491 merges and remove the label for the JIRA username field.

robbkidd:
Updated PR to _not_ include a label for JIRA field because #1491 will be merging first and removes that field. Now we have the option to rebase or merge because the two branches should not conflict.

nellshamrell:
:shipit:

PR #1488, opened 2016-12-02
Remove cla content

nellshamrell:
Removes CLA content from the manage profile views.


nellshamrell:
This is now ready for review.

src/supermarket/app/views/profile/edit.html.erb
```diff
@@ -55,7 +52,6 @@
 
       <% if ROLLOUT.active?(:github) %>
         <div class="content" id="github-accounts">
-          <p>You must have at least one linked GitHub account in order for us to match your contributions on GitHub with your signed Contributors License Agreement.</p>
```
robbkidd: With this text removed, the GitHub account tab now appears as:

screen shot 2016-12-06 at 1 13 34 pm

The button seems so lonely. And I'm not sure why I would want to link my GitHub account. Instead of removing text  entirely, what do you think about replacing it with something along the following?

> Linking your account lets other cookbook shoppers find you on GitHub for feedback, collaboration, and kudos. The link requests access only to your GitHub account's public information.

nellshamrell: I think that is a fantastic idea!

nellshamrell: Done!

robbkidd: 👍

PR #1490, opened 2016-12-02
Modify key wording

nellshamrell:
Fixes #1464

PR #1491, opened 2016-12-06
stop displaying JIRA username

### LINKED ISSUE #1484: Remove JIRA name from Supermarket profile

robbkidd:
The Chef community does not use JIRA to track issues anymore, so asking
for and displaying a JIRA username is confusing.

Edit profile feature test changed to set the previously-untested Twitter
username which is still a field on the form.

Fixes #1484


robbkidd:
More Manage Profile cleanup. I recommend this goes in after #1490 because the update to the contents of the Keys tab fixes a real problem vs this cleanup change.

nellshamrell:
:shipit:

PR #1492, opened 2016-12-07
Use chef_nginx instead of nginx

tas50:
Nginx is no longer being maintained. Depend on the nginx cookbook instead.


robbkidd:
This one would have been good, but has become overcome by the events of #1501 where we removed the dependency on any nginx cookbook by copying the one template we were using into our internal omnibus cookbook.

PR #1493, opened 2016-12-07
Renovate fieri code

### LINKED ISSUE #1479: Refactor Fieri Metrics which use Cookbooks JSON api response
We have a couple of metrics run through Fieri which query the cookbook api response (along with one that queries the cookbook version api response).  Right now we query the cookbook api for EACH of these metrics, we should consider refactoring so that we query the cookbook api only once, then pass the response to each of the metrics.

nellshamrell:
Fixes #1479

### DIFF
--- a/src/fieri/app/controllers/fieri/jobs_controller.rb
+++ b/src/fieri/app/controllers/fieri/jobs_controller.rb
@@ -3,11 +3,7 @@ require_dependency 'fieri/application_controller'
 module Fieri
   class JobsController < ApplicationController
     def create
-      CollaboratorWorker.perform_async(job_params[:cookbook_name])
-      FoodcriticWorker.perform_async(job_params)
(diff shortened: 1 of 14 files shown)


robbkidd:
I am wary of making HTTP calls _during_ an HTTP call, so I'm thinking the controller is not the place to make the calls back to Supermarket proper.

I agree that we can use something that makes the calls back to Supermarket once for information needed by multiple metric workers. I think that something maybe ought to be a worker itself, like a `MetricRunner` or something.

`JobController` as a controller hands off the params it was given to an async job for `MetricRunner`. Then `MetricRunner`'s perform makes the calls back to Supermarket and schedules the individual metric Workers with the parameters they need.

nellshamrell:
I am open to that.  Implementing.

nellshamrell:
This is again ready for review.

nellshamrell:
Changes have been made, this is ready for review.

PR #1494, opened 2016-12-07
pin to version of runit that worked

robbkidd:
We can't leave this, but it unblocked the pipeline while the problem is being worked.


nellshamrell:
👍

robbkidd:
Noting here that runit was fixed in v3.0.3 and we declared that as a minimum version in #1495.

PR #1495, opened 2016-12-07
set runit cookbook version equal or greater than the version a bug fix appears in

nellshamrell:


robbkidd:
I have tested this change with a locally-built omnibus package and it fixes the nginx running with the right config problem.

I'm torn between:

* unpinning which should work for CI because we don't check in `Berksfile.lock`, but could have unexpected results in local builds
* setting runit version to `>= 3.0.3` to ensure a build (local or CI) uses a version without the bug

nellshamrell:
Let's go ahead and pin to >= 3.0.3, since we know that will avoid the bug. Making that change now.

robbkidd:
`git squoosh`?

nellshamrell:
Squoosh done

PR #1500, opened 2016-12-14
Refactor supermarket quality metrics

### LINKED ISSUE #1478: Refactor API::CookbookVersions controller around metrics collection
The API::CookbookVersions controller is quickly becoming unwieldy, especially as we continue to add metrics.  We need to re-architect/restructure it to avoid continuing to accumulate technical debt and avoid making it hard to maintain.

nellshamrell:
Fixes #1478


nellshamrell:
This is now ready for review.

src/supermarket/app/controllers/api/v1/quality_metrics_controller.rb
```diff
@@ -0,0 +1,171 @@
...
+    params.require(:fieri_key)
+    params.require(:cookbook_name)
+    params.require(:cookbook_version)
+    params.require(:foodcritic_failure)
+  end
+
+  def require_collaborator_params
+    params.require(:cookbook_name)
+    params.require(:collaborator_failure)
+    params.require(:collaborator_feedback)
```
robbkidd: I am curious as to why `:fieri_key` is not required for the `collaborator_evaluation` endpoint. I know the parameter gets used by the `check_authorization` filter before the controller action is called. I don't think anything we do _in_ the controller action uses the `:fieri_key`. Maybe we don't need to include `:fiery_key` param in these `require_thingie_params` filters?

src/supermarket/app/controllers/api/v1/quality_metrics_controller.rb
```diff
@@ -0,0 +1,171 @@
...
+  #
+  # If the +CookbookVersion+ does not exist, render a 404 not_found.
+  #
+  # If the request is unauthorized, render unauthorized.
+  #
+  # This endpoint expects +cookbook_name+, +cookbook_version+,
+  # +publish_failure+, +publish_feedback+, and +fieri_key+.
+  #
+  def publish_evaluation
+    create_metric(
```
robbkidd: Are there no required parameters for this evaluation?

PR #1501, opened 2016-12-15
Optional logging of X-Forwarded-For, ease parsing with custom nginx log format

robbkidd:
This change pulls in the `nginx.conf` template from the version of the nginx cookbook we've been pinned to for a long while. This lets us remove the dependency on the actual nginx cookbook because now we own our templates and we get our production nginx binaries from the omnibus build.

There is a lot of change here, but each commit is scoped to a particular change with a reason recorded in the commit message. Walking the commits will probably be the clearest way to review.


robbkidd:
Sending over to the pipeline to ready it for review and approval today so we can get it on Staging for to parse the logs.

omnibus/cookbooks/omnibus-supermarket/templates/default/nginx.conf.erb
```diff
@@ -0,0 +1,83 @@
...
+}
+
+http {
+  <% if @nginx['log_x_forwarded_for'] -%>
+  log_format supermarket '$http_x_forwarded_for - $remote_user [$time_iso8601]  '
+  <% else -%>
+  log_format supermarket '$remote_addr - $remote_user [$time_iso8601]  '
+  <% end -%>
+                     '"$request" $status $request_length "$request_time" $body_bytes_sent '
+                     '"$http_referer" "$http_user_agent"';
```
joshbrand: this likes it will do the thing! 👍

Issue #1201, opened 2016-01-08
Add chef_version and ohai_version to database tables


lamont-granquist:
This is the task to extract the chef_version and ohai_version out of the uploaded cookbook metadata and to get them into a table.

Displaying the data in the UI comes later (and I actually don't super care about it as a priority)

Getting the chef_version and ohai_version into /universe also comes later (this is what I want to implement on top of this issue).

robbkidd:
@lamont-granquist Can you add some examples of the various forms these versions may come in via cookbook metadata? I recall something about strings or arrays of strings.

lamont-granquist:
PR that added them to chef/chef:  https://github.com/chef/chef/pull/4081

lamont-granquist:
Here's an example of the sub-JSON:

https://github.com/chef/chef/blob/a75ce7ce4c5ee9565f536430244a37bc33cd4e59/lib/chef/cookbook/metadata.rb#L526

nellshamrell:
@lamont-granquist: what am I looking for?  I see that a cookbook itself can have multiple chef versions and multiple ohai versions, but I believe a cookbook VERSION can only have one chef version and one ohai version.  Is there something that needs to change?

nellshamrell:
OH wait a minute.  It looks like people can specify that a cookbook works with multiple versions of Chef in their metadata.rb.  Therefore, I believe the chef_version and ohai_version attributes should be on the Cookbook object, rather than the CookbookVersion object and should allow arrays of strings.  Any objections to this?

lamont-granquist:
i think i ninja answered you?

also note that they key will be "ohai_versions" and "chef_versions" in the plural.

nellshamrell:
Right.  So it looks like the changes that need to be made are:
1) Rename ohai_version and chef_version to ohai_versions and chef_versions
2) Move ohai_versions and chef_versions to be attributes on the Cookbook object, rather than the CookbookVersion object
3) Add in the ability to handle an array of strings.

Is this correct?

lamont-granquist:
I'm not sure what Cookbook and CookbookVersion mean to you?  Each version of a cookbook will have different chef_versions and ohai_versions.

lamont-granquist:
So for a 'foo' cookbook:

foo 0.0.0 may not have anything specified
foo 1.0.0 may add something like chef_versions: [ [ "~> 11.14.0" ] ]
[...]

PR #1503, opened 2016-12-22
Add chef ohai

### LINKED ISSUE #1379: Display {chef|ohai}_version for a cookbook
We now include a chef_version metadata item on later releases of Chef 12.  If that data is present we should surface it on the Supermarket. It would be really helpful to allow people to [...] see it on the cookbook pages.

@robbkidd updated: Moved the search aspect of this feature to #1407.

nellshamrell:
Fixes #1379 

Here is what I have come up with for displaying the versions of Chef and Ohai that a cookbook version supports.  I welcome any input on the view, etc.

screen shot 2016-12-21 at 4 39 41 pm

(I will squash the commits when this is finalized)

### DIFF
--- a/src/supermarket/app/helpers/cookbook_versions_helper.rb
+++ b/src/supermarket/app/helpers/cookbook_versions_helper.rb
@@ -66,4 +66,21 @@ module CookbookVersionsHelper
       content
     end
   end
+
+  def versions_string(versions)
+    versions_string = ''
+
+    versions.each do |version_set|
+      versions_string += "(#{combined_set(version_set)})"
+      versions_string += ' OR ' unless version_set == versions.last
+    end
+
+    versions_string
+  end
+
+  private
+
+  def combined_set(set)
+    set.join(' AND ')
+  end
 end
--- a/src/supermarket/app/views/cookbooks/_sidebar.html.erb
+++ b/src/supermarket/app/views/cookbooks/_sidebar.html.erb
@@ -90,6 +90,16 @@
     <h4><i class="fa fa-key"></i> License</h4>
     <p><%= version.license %></p>
 
+    <% if version.chef_versions.present? %>
+      <h4><i class="fa fa-key"></i> Chef Versions</h4>
+      <p><%= versions_string(version.chef_versions) %></p>
+    <% end %>
+
+    <% if version.ohai_versions.present? %>
+      <h4><i class="fa fa-key"></i> Ohai Versions</h4>
+      <p><%= versions_string(version.ohai_versions) %></p>
+    <% end %>
+
     <%= link_to "Download Cookbook", { action: 'download' }, class: 'button secondary radius expand button_download_cookbook' %>
   </div>
 </div>
--- a/src/supermarket/spec/helpers/cookbook_versions_helper_spec.rb
+++ b/src/supermarket/spec/helpers/cookbook_versions_helper_spec.rb
@@ -24,4 +24,26 @@ describe CookbookVersionsHelper do
       expect(helper.safe_updated_at(nil)).to be <= Time.zone.now
     end
   end
+
+  describe 'versions_string' do
+    let(:chef_versions) { [['>= 12.4.1', '< 12.4.2'], ['> 11.2.3', '<= 12.4.3']] }
+    let(:string) { helper.versions_string(chef_versions) }
+
+    context 'combining the inner array elements' do
+      it 'combines them with AND' do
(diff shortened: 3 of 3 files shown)


nellshamrell:
This is ready for review.

tas50:
@nellshamrell What metadata string produced the example text in the screenshot? Most of the Chef versions are pretty basic ~ 12.0 or >= 12.0 sort of things. Those will display as expected right?

nellshamrell:
That example was provided by Lamont in #1379

I expect the more basic version numbers will work fine as well, but I will verify.

nellshamrell:
@lamont-granquist ^

nellshamrell:
Pardon me, I got that example from @lamont-granquist in #1201, not #1379

src/supermarket/app/helpers/cookbook_versions_helper.rb
```diff
@@ -66,4 +66,29 @@ def render_document(content, extension)
       content
     end
   end
+
+  def chef_versions_string(chef_versions)
+    versions_string(chef_versions)
+  end
+
+  def ohai_versions_string(ohai_versions)
```
robbkidd: Yey, helpers! I wonder, though, if `versions_string()` could be promoted to public and used directly. There is currently no difference in rendering a collection of version constraint ranges.

src/supermarket/spec/helpers/cookbook_versions_helper_spec.rb
```diff
@@ -24,4 +24,48 @@
       expect(helper.safe_updated_at(nil)).to be <= Time.zone.now
     end
   end
+
+  describe 'chef_versions_string' do
+    let(:chef_versions) { [['12.4.1', '12.4.2'], ['11.2.3', '12.4.3']] }
```
robbkidd: I think each of these example versions ought to include an operator for consistency with the expect usage of the version constraints. 

I agree that the implementation is correct, though. I tossed the following into a working copy of this PR and things were green:

```ruby
    context 'with some operators' do
      it 'handles a simple range' do
        versions = [ [">= 11.14.5", "< 12.0.0"] ]
        expected_output = "(>= 11.14.5 AND < 12.0.0)"
        expect(helper.chef_versions_string(versions)).to eq(expected_output)
      end

      it 'handles what Lamont says are gnarly ranges' do
        versions = [
          [ ">= 11.14.5", "< 12.0.0" ],
          [ ">= 12.4.2", "< 13.0.0" ]
        ]
        expected_output = "(>= 11.14.5 AND < 12.0.0) OR (>= 12.4.2 AND < 13.0.0)"
        expect(helper.chef_versions_string(versions)).to eq(expected_output)
      end
    end
```

nellshamrell: Will add that in.

robbkidd: Looking again, I'm not sure my bit ought to be added as much as operators should be included in all version constraints appearing in the tests.

nellshamrell: I don't have a strong feeling either way.  I'll add in a combination of them.

nellshamrell: Ah, I see your above comment, will add operators to each example.

nellshamrell:
Changes have been made.

I tested defining a cookbook version's chef_versions attribute as only one version.  If only one chef or ohai version is supported, it will need to be entered like this:

chef_versions: [[">= 12.4.1"]] 

It will need to be an array within an array.

nellshamrell:
It should also be noted that there is currently no way of adding chef_versions or ohai_versions attributes to a cookbook version other than through the console.

src/supermarket/app/views/cookbooks/_sidebar.html.erb
```diff
@@ -90,6 +90,16 @@
     <h4><i class="fa fa-key"></i> License</h4>
     <p><%= version.license %></p>
 
+    <% if version.chef_versions.present? %>
+      <h4><i class="fa fa-key"></i> Chef Versions</h4>
```
robbkidd: I'm looking for a font-awesome icon that might be more appropriate here than 🗝

robbkidd: We can find something more appropriate later, maybe after upgrading the version of font-awesome we use.

PR #1513, opened 2017-01-05
implement QM006 - cookbook declares at least one supported platform

### LINKED ISSUE #1476: Implement  "Cookbook lists at least one platform it supports" metric
https://github.com/chef-cookbooks/cookbook-quality-metrics/blob/master/quality-metrics/qm-006-supports.md

robbkidd:
Implementing a metric for [QM006](https://github.com/chef-cookbooks/cookbook-quality-metrics/blob/52be6b20a891e0b2f0915df2ec58beb5588141b5/quality-metrics/qm-006-supports.md)

Closes #1476

TODO:
- [x] need an evaluation endpoint to receive the feedback


src/fieri/app/models/supported_platforms_worker.rb
```diff
@@ -0,0 +1,49 @@
...
+      cookbook_version: version_info['version'],
+      supported_platforms_failure: supported_platforms_failure?,
+      supported_platforms_feedback: supported_platforms_feedback
+    )
+
+    return if response.is_a? Net::HTTPSuccess
+    raise "Unable to POST Supported Platforms Evaluation of #{cookbook_name}: #{response.code} #{response.message}"
+  rescue => e
+    log_error(e)
+  end
```
robbkidd: I modeled this rescue'n'log pattern from the `FoodcriticWorker`. So, yay for logged errors, but as-is the rescue swallows the exception and Sidekiq thinks the job succeeded and won't retry. I'm going for a "_yes_ this is how we want to make workers" goal here, if only on one of them and I think I want retries.

Thoughts about whether that's how we want to keep it or whether we should spec the exceptions to be raised so the job is attempted again?

nellshamrell:
👍

PR #1519, opened 2017-01-06
include the published_at in API info for cookbook version

### LINKED ISSUE #1515: Add cookbook version release data to the supermarket API
It would be incredibly useful to have the ability to poll the release data of a version via the supermarket API. This allows us to figure out how many releases we've done in the last year. Instead we have to do this by checking out every repo and pulling every tag.

robbkidd:
Went with "published_at" based on created_at time.

Looks a bit like:

```json
{
  "license": "blargle",
  "tarball_file_size": 1698,
  "version": "0.1.0",
  "published_at": "2017-01-06T19:05:51Z",
  "average_rating": null,
  "cookbook": "http://localhost:3000/api/v1/cookbooks/blargle",
  "file": "http://localhost:3000/api/v1/cookbooks/blargle/versions/0.1.0/download",
  "quality_metrics": []
}
```

Closes #1515

### DIFF
--- a/src/supermarket/app/views/api/v1/cookbook_versions/_cookbook_version.json.jbuilder
+++ b/src/supermarket/app/views/api/v1/cookbook_versions/_cookbook_version.json.jbuilder
@@ -1,6 +1,7 @@
 json.license cookbook_version.license
 json.tarball_file_size cookbook_version.tarball_file_size
 json.version cookbook_version.version
+json.published_at cookbook_version.created_at.iso8601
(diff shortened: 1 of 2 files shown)


nellshamrell:
I don't forsee any issues with this.  The only way I could see an error is if a cookbook_version does not have an updated_at value for some reason....which I don't offhand think is the case for any cookbooks.

nellshamrell:
👍

PR #1521, opened 2017-01-18
update omniauth to fix security issue

robbkidd:
Described in omniauth/omniauth#867 and related to chef/chef-server#1042


nellshamrell:
👍

PR #1524, opened 2017-01-19
adds constraint to the openssl cookbook to avoid breakage in an earli…

nellshamrell:
…er version


tas50:
👍

PR #1525, opened 2017-01-23
adds passing message to publish worker results

nellshamrell:

### DIFF
--- a/src/fieri/app/models/publish_worker.rb
+++ b/src/fieri/app/models/publish_worker.rb
@@ -25,6 +25,10 @@ class PublishWorker
       publish_feedback += "#{cookbook_name} is up for adoption\n"
     end
 
+    unless failure == true
+      publish_feedback += "#{cookbook_name} passed the publish metric"
+    end
+
     Net::HTTP.post_form(
       URI.parse("#{ENV['FIERI_SUPERMARKET_ENDPOINT']}/api/v1/quality_metrics/publish_evaluation"),
       fieri_key: ENV['FIERI_KEY'],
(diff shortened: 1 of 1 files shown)


src/fieri/app/models/publish_worker.rb
```diff
@@ -25,6 +25,10 @@ def perform(cookbook_json, cookbook_name)
       publish_feedback += "#{cookbook_name} is up for adoption\n"
     end
 
+    unless failure == true
```
robbkidd: What do you think about only `unless failure` here?

src/fieri/app/models/publish_worker.rb
```diff
@@ -25,6 +25,10 @@ def perform(cookbook_json, cookbook_name)
...
+      publish_feedback += "#{cookbook_name} passed the publish metric"
```
robbkidd: 🤔 While it's not likely that we would end up here with something other than `''` as the value for feedback, maybe an `=` here instead of an append `+=`?

I'm picturing a "no failure? Okay, THIS IS ALL OF YOUR FEEDBACK, GOD JORB" sort of thing.

nellshamrell: Agreed, will change that to an =

nellshamrell: Done!

src/fieri/app/models/publish_worker.rb
nellshamrell: Done!

robbkidd:
Yep!

PR #1540, opened 2017-02-02
🚧 WIP: metric rerunner 🚧

### LINKED ISSUE #1530: an operator can schedule a rerun of qual metrics for latest versions of cookbooks

robbkidd:
Super WIPy.

Will eventually close #1530


nellshamrell:
This initial architecture looks fine and is an outstanding use of a rails service object.  I approve!

src/supermarket/app/services/run_quality_metrics.rb
```diff
@@ -0,0 +1,18 @@
+class RunQualityMetrics
+  attr_reader :cookbook
+
+  def initialize(cookbook)
+    @cookbook = cookbook
+  end
+
+  def self.all_the_latest
+    Cookbook.all.each do |cookbook|
+      new(cookbook).call
```
robbkidd: This could be a doozy of an N+1. Like, around 3,200+1 on public Supermarket.

robbkidd: Because we really only need the `id` of the latest version of a cookbook, this almost gets us there:

```
CookbookVersion.where('
  cookbook_versions.created_at = (
    SELECT MAX(latest.created_at) 
    FROM cookbook_versions AS latest 
    WHERE latest.cookbook_id = cookbook_versions.cookbook_id
  )
').pluck(:id)
```

It optimizes around the assumption that the latest cookbook version is the most recently created/published/shared/uploaded. This is not _strictly_ true—our `cookbook.latest_cookbook_version` is implemented as a sort on SemVer—but date-based-latest is probably true enough for a ReRunMetricsOnLatest so that we can make one beefy call to the database and receive the list of CookbookVersion ids.

There's no getting around the need to schedule ~3,200 fieri notifications, though. 😟

gif-keyboard-10193540256711913767

robbkidd:
At this point, the UX of this feature would be:

```
> sudo -u supermarket supermarket-ctl console
console> RunQualityMetrics.all_the_latest
=> 'cool-job-id,bro'

console> RunQualityMetrics.on_latest 'apache2'
=> 'yet-another-job-id'

console> RunQualityMetrics.on_version 'apache2', '1.2.3'
=> 'this-is-your-job-id-there-are-many-like-it-but-this-one-is-yours'
```

How much deeper shall we go? Rake tasks for each and -ctl commands to call them?

nellshamrell:
I would support adding rake tasks, then adding ctl commands that call those rake tasks.  That is the way the make_admin ctl command works and it seems to be a decent pattern.

robbkidd:
Okeydoke. I'll get on rake tasks and -ctl commands to run them.

Also, I'll probably add a little error checking and reporting to the two methods that take user input—`on_latest` and `on_version`—to feedback issues with typoed names or versions that don't exist.

nellshamrell:
This looks beautiful so far.

robbkidd:
🌈

robbkidd:
Now that I've got the service class returning messages about success or failure, I sense the need for maybe returning _tuples_ to pass along happy/sad state with the message.

Returns like mebbe,

```[:ok, "Scheduled quality metric evaluations for 13 cookbook(s)."]
[:error, "The nope cookbook was not found."]
```

... so that the consuming thing (rake task, future API controller endpoint) can sort out displaying happy/sad (like STDERR vs STDOUT, or which alert to flash in a view) without parsing the text or handling an exception (because it's not exceptional that users give us bad input).

nellshamrell:
Tuples sound like a good structure choice.

robbkidd:
Updated with tuplage. _Now_ on to rake tasks.

robbkidd:
@nellshamrell Ready for review.

PR #1543, opened 2017-02-13
make the API item limit configurable

### LINKED ISSUE #1542: Make API items limit number a configurable value
We are working on an integration with a partner where they need the ability to batch download cookbooks. In order to do so we would like to raise the limit on the # of cookbooks downloadable at one time from 100 to something like 10000.

As per discussion with @robbkidd it appears that Supermarket [hardcodes the limit to 100](https://github.com/chef/supermarket/commit/6c93304f22e29b2d6e0606131ce07a853fe59015#diff-f9df768289fd6bf1c359ddab5b920e25R5) right now.

robbkidd:
Use `API_ITEM_LIMIT` environment variable to set the upper limit on the number of items to return for API requests to a Supermarket instance.

Defaults to the previously hard-coded value of 100 if the env var is not present. Also, the cookbook internal to the omnibus package has the same default set in the section of the attributes that get set as env vars.

Closes #1542

### DIFF
--- a/src/supermarket/app/controllers/api/v1_controller.rb
+++ b/src/supermarket/app/controllers/api/v1_controller.rb
@@ -46,7 +46,7 @@ class Api::V1Controller < ApplicationController
   #
   def init_params
     @start = params.fetch(:start, 0).to_i
-    @items = [params.fetch(:items, 10).to_i, 100].min
+    @items = [params.fetch(:items, 10).to_i, item_limit].min
 
     if @start < 0 || @items < 0
       return error(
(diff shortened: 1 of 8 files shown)


src/supermarket/app/controllers/api/v1_controller.rb
```diff
@@ -46,7 +46,7 @@ def error(body, status = 400)
   #
   def init_params
     @start = params.fetch(:start, 0).to_i
-    @items = [params.fetch(:items, 10).to_i, 100].min
+    @items = [params.fetch(:items, 10).to_i, item_limit].min
```
robbkidd: There's a part of me that thinks that the API should respond with a 4xx error when an API request has an `item` value greater than the limit. Then the requestor would get a clear "no soup for you" message instead of wondering why their request did not return their desired number of items.

nathenharvey: agreed but can happen after we release.  Can you open an issue for that before this gets released?

robbkidd: Opened #1544 to return an error when an item count is requested greater than the item limit.

robbkidd:
Submitting to the pipeline for a build to test further.

PR #1549, opened 2017-02-23
really, really disable the datadog tracer

robbkidd:
The tracer at the current version still tries to make connection attempts to the Datadog agent even when disabled. To disable those attempts and the log spam errors it generates, we'll only require the gem when we have enabled the tracer.

See datadog/dd-trace-rb#53 issue about the symptom.


robbkidd:
👍  Rails and Sidekiq services are happier with this style feature flipper. No log spam, so presumably no extraneous connection attempts for the disabled agent.

PR #1551, opened 2017-02-24
Contributing file QM

### LINKED ISSUE #1532: QM011 - Contributing doc

nellshamrell:
Closes #1532

### DIFF
--- a/src/fieri/app/models/contributing_file_worker.rb
+++ b/src/fieri/app/models/contributing_file_worker.rb
@@ -2,18 +2,18 @@ require 'sidekiq'
 require 'net/http'
 require 'octokit'
 
-class ContributorFileWorker
+class ContributingFileWorker
   include ::Sidekiq::Worker
 
   def perform(cookbook_json)
     evaluate_result = evaluate(cookbook_json)
 
     Net::HTTP.post_form(
-      URI.parse("#{ENV['FIERI_SUPERMARKET_ENDPOINT']}/api/v1/quality_metrics/contributor_file_evaluation"),
+      URI.parse("#{ENV['FIERI_SUPERMARKET_ENDPOINT']}/api/v1/quality_metrics/contributing_file_evaluation"),
       fieri_key: ENV['FIERI_KEY'],
       cookbook_name: cookbook_name(cookbook_json),
-      contributor_file_failure: evaluate_result,
-      contributor_file_feedback: give_feedback(evaluate_result)
+      contributing_file_failure: evaluate_result,
+      contributing_file_feedback: give_feedback(evaluate_result)
     )
   end
 
@@ -49,11 +49,11 @@ class ContributorFileWorker
   def give_feedback(failure_result)
     if failure_result
       I18n.t(
-        'quality_metrics.contributor_file.failure'
+        'quality_metrics.contributing_file.failure'
       )
     else
       I18n.t(
-        'quality_metrics.contributor_file.success'
+        'quality_metrics.contributing_file.success'
       )
     end
   end
--- a/src/supermarket/app/models/quality_metric.rb
+++ b/src/supermarket/app/models/quality_metric.rb
@@ -27,6 +27,6 @@ class QualityMetric < ActiveRecord::Base
   end
 
   def self.contributing_file_metric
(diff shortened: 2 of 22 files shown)


nellshamrell:
@robbkidd this is now ready for review, though I'm wondering if it would be worth the effort to rename every variant of "Contributor File" to "Contributing File."

robbkidd:
I think "contributing" makes more sense. Probably with it for future sanity.

nellshamrell:
Renamed Contributor File Metric to Contributing File Metric, @robbkidd this is ready for review.

src/fieri/app/models/contributing_file_worker.rb
```diff
@@ -0,0 +1,66 @@
...
+    # http://rubular.com/r/wHJrs02iLs
+    cookbook_json.match(%r{(?<=github.com\/)[\w-]+\/[\w-]+}).to_s
+  end
+
+  def evaluate(cookbook_json)
+    repo = source_user_repo(cookbook_json)
+
+    if repo.empty?
+      # if no match for repo from #source_user_repo, fails metric
+      return true
```
robbkidd: Is there some more feedback we can provide about why this metric failed?

```
Cookbook does not declare a source repository. Cannot check for presence of CONTRIBUTING document.
Source repo is not a GitHub repo. (only supports GitHub at the moment)
Source repo does not exist.
```

src/fieri/app/models/contributing_file_worker.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    begin
+      octokit_client.contents(repo, path: 'CONTRIBUTING.md')
```
robbkidd: I see that [QM011](https://github.com/chef-cookbooks/cookbook-quality-metrics/blob/f45c46c5a24efc397a2749f16fb8bf5549014ff0/quality-metrics/qm-011-contributing_doc.md) states specifically `CONTRIBUTING.md`, but _must_ a document be in markdown? Maybe that's a question for an adjustment to QM011, so `.md` is OK for this first version of the metric.

src/supermarket/app/models/quality_metric.rb
```diff
@@ -25,4 +25,8 @@ def self.license_metric
   def self.supported_platforms_metric
     QualityMetric.where(name: 'Supported Platforms').first_or_create!(admin_only: true)
   end
+
+  def self.contributing_file_metric
+    QualityMetric.find_by(name: 'Contributing File')
```
robbkidd: What are your thoughts around this finder not following the `QM.where(name:).first_or_create!()` pattern of the others?

In the last round of metric work, I changed these so that both the seeds and the class finder methods would ensure these fixture-y records existed. I'm open to dialing the creation back to only seeds, but I suggest these finder methods should be consistent with each other.

nellshamrell: Ah!  My mistake!  Will get that corrected.

src/fieri/app/models/contributing_file_worker.rb
nellshamrell: Let's leave it as a potential adjustment for QM011

src/fieri/app/models/contributing_file_worker.rb
nellshamrell: Beyond what is returned in the give_feedback? method? That is currently:

'Failure: To pass this metric, your cookbook metadata must include a source url, the source url must be in the form of http://github.com/user    /repo, and your repo must contain a CONTRIBUTING.md file'

src/supermarket/app/models/quality_metric.rb
nellshamrell: Fixed!

src/fieri/app/models/contributing_file_worker.rb
robbkidd: Dur ... I should have come back to this comment of mine after I got to the code change with the I18n'd messages. Right. Disregard.

nellshamrell:
Anything else need review on this?

robbkidd:
I don't think so. Thumbing it.

PR #1552, opened 2017-02-25
make sidekiq optionally connect to an alternate redis

robbkidd:
Adds an initializer for Sidekiq to configure which Redis instance to use for job queuing. Defaults to a common single `REDIS_URL` that will be shared with Rails caching and Rollout feature flag checking. Shared is less than ideal in high traffic production environments, so this adds the option to set a different Redis URL for Sidekiq.

https://github.com/mperham/sidekiq/wiki/Using-Redis#using-an-initializer

TODO:
- [x] decide if `REDIS_JOBQ_URL` is too silly a name and, if so, pick something else
- [x] set `default['supermarket']['redis_jobq_url']` to something sensible in the omnibus' cookbook (probably match `redis_url` but with a different Redis DB id)

### DIFF
--- a/omnibus/cookbooks/omnibus-supermarket/attributes/default.rb
+++ b/omnibus/cookbooks/omnibus-supermarket/attributes/default.rb
@@ -332,7 +332,8 @@ default['supermarket']['port'] = node['supermarket']['nginx']['force_ssl'] ? nod
 default['supermarket']['protocol'] = node['supermarket']['nginx']['force_ssl'] ? 'https' : 'http'
 default['supermarket']['pubsubhubbub_callback_url'] = nil
 default['supermarket']['pubsubhubbub_secret'] = nil
-default['supermarket']['redis_url'] = "redis://#{node['supermarket']['redis']['bind']}:#{node['supermarket']['redis']['port']}/0/supermarket"
+default['supermarket']['redis_url'] = 'redis://127.0.0.1:16379/0/supermarket'
+default['supermarket']['redis_jobq_url'] = nil
 default['supermarket']['sentry_url'] = nil
 default['supermarket']['api_item_limit'] = 100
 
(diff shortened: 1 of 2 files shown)


nellshamrell:
👍

robbkidd:
Thanks! Working on adding the attribute to the omnibus' internal cookbook now.

omnibus/cookbooks/omnibus-supermarket/attributes/default.rb
```diff
@@ -332,7 +332,8 @@
 default['supermarket']['protocol'] = node['supermarket']['nginx']['force_ssl'] ? 'https' : 'http'
 default['supermarket']['pubsubhubbub_callback_url'] = nil
 default['supermarket']['pubsubhubbub_secret'] = nil
-default['supermarket']['redis_url'] = "redis://#{node['supermarket']['redis']['bind']}:#{node['supermarket']['redis']['port']}/0/supermarket"
+default['supermarket']['redis_url'] = 'redis://127.0.0.1:16379/0/supermarket'
```
robbkidd: Took the opportunity to update this one attribute related to the change to remove the interpolation of default values. These interpolations are not evaluated later to pick up overrides which is what many humans reading these attributes think happens.

I'll also note here that I've observed that the `/supermarket` portion of the Redis connection URL does nothing; keys go into the numbered database at the "root" not under a "supermarket" namespace. And [the scheme for a Redis URI](https://www.iana.org/assignments/uri-schemes/prov/redis) has the path ending in a sole, lonely positive integer to be followed by optional parameters.

Some other time, I'll pull the `/supermarket` portions out of the defaults. I did not include that in this change to minimize differences.

robbkidd:
Attribute added. I'm going to accept that thumb and push this change into the pipeline so we can try it out on Staging.

PR #1559, opened 2017-03-07
Adding testing file QM

nellshamrell:


src/fieri/app/models/metrics_runner.rb
```diff
@@ -24,4 +24,8 @@ def cookbook_api_response(params)
   def cookbook_version_api_response(params)
     SupermarketApiRunner.new.cookbook_version_api_response(params['cookbook_name'], params['cookbook_version'])
   end
+
+  def source_repo(cookbook_json)
+    cookbook_json.match(%r{(?<=github.com\/)[\w-]+\/[\w-]+}).to_s
```
robbkidd: I get the feeling that this makes `MetricsRunner` have to know too much about implementation details of the various metric workers. Maybe we could consider a `SourceRepoMetric` class from which both `ContributingFileWorker` and the soon-to-exist `TestingFileWorker` could descend. That parent class could contain the logic for looking up the source repository URL from `cookbook_json` and could provide a shared, private `octokit_client` to hide the details of how to determine "source contains a thingie."

Speaking of looking up the source repo URL, I'd prefer a JSON.parse-gimmie-['source_url'] over a capturing regex treating the JSON response like a string.

nellshamrell: On your first point, I can see that, having something like a GithubMetric class that the Contributing Metric and Testing Metric both inherit from.

On your second point, I will check into doing that.

robbkidd: I was initially thinking `GitHubMetric` parent class, but while re-reading the metric definitions, I considered a more generic name like `SourceRepoMetric` (which I agree is still not good) about "checking that something exists in the source repo" and that GitHub just happens to be the first git service we'll support.

nellshamrell: I ended up going with SourceRepoMetric - I think it's the best we can do for now.

As for your second, I can access the full github url with 
```
JSON.parse(cookbook_json)['source_url']
```
which will return something like "http://github.com/some_user/some_repo" , but I still will need to parse it with a regex because the github url will only accept "some_user/some_repo" when using the contents api.

nellshamrell:
@robbkidd this is ready for review whenever you have time.

src/fieri/app/models/metrics_runner.rb
robbkidd: > I still will need to parse it with a regex because the github url will only accept "some_user/some_repo" when using the contents api.

Ugh. My bad. I missed that nuance to the string needed to Do The Thing™.

facepalm

So that we don't treat a structured JSON response like a big string, I _think_ I still like ...

```
JSON.parse(cookbook_json)['source_url'].match(%r{(?<=github.com\/)[\w-]+\/[\w-]+}).to_s
```

PR #1567, opened 2017-03-15
upgrade fieri's foodcritic to 10.0

robbkidd:
foodcritic 10 switched to showing progress by default, so added the `--no-progress` switch to the criticizer's arguments.

TODO:
- [x] set new desired version in `src/fieri/fieri.gemspec`
- [x] `bundle install` in both `src/fieri` and `src/supermarket` to get new version officially pinned in the locks
- [ ] adjust the tags enabled? (for [testing](https://github.com/chef/supermarket/blob/master/src/fieri/.env#L2) and for [production default](https://github.com/chef/supermarket/blob/master/src/supermarket/.env#L24))


robbkidd:
@tas50 Are there adjustments to the default tags enabled we should make with foodcritic 10? 

Currently, they are `metadata,correctness ~FC031 ~FC04`.

tas50:
It may be useful at add supermarket, portability, and deprecated tags

robbkidd:
So that would become `metadata,correctness,supermarket,portability,deprecated ~FC031 ~FC04`?

What might be an appropriate set to use for private Supermarkets?

tas50:
The same. If you're large enough to need a Supermarket you probably want people to include a readme and properly define the maintainer

PR #1570, opened 2017-03-16
QM012 - cookbook contains no binaries

### LINKED ISSUE #1535: QM012 - no binaries

robbkidd:
:do_not_litter: [no binaries](https://github.com/chef-cookbooks/cookbook-quality-metrics/blob/master/quality-metrics/qm-012-binaries.md) :no_good:

Closes #1535 

TODO:
- [x] an omnibus software definition for `file` (includes `libmagic`)
- [x] bring in `ruby-filemagic` gem for testing whether a file is a binary
- [x] teach CookbookArtifact how to test a tarball for binaries
- [x] include a file size check for encoding shenanigans
- [x] add a NoBinariesWorker to produce the metric feedback
- [x] add NoBinaries metric result type and API endpoint to receive feedback
- [ ] judicious testing of the omnibus build including `libmagic` and fieri using it properly


robbkidd:
@nellshamrell Partial progress on one of the options for detecting binaries. To review the options:

+ 👉  in Ruby using [ruby-filemagic](https://github.com/blackwinter/ruby-filemagic) and the libmagic C library
    + requires a `file` omnibus software definition (which includes the `libmagic` library) - liberally cribbed in this PR from the [`core/file` Habitat plan](https://github.com/habitat-sh/core-plans/blob/d611be73c0314d72eb4cc826c7a0fc54c64b81d9/file/plan.sh)
+ in Ruby using [charlock-holmes](https://github.com/brianmario/charlock_holmes) and the [ICU](http://site.icu-project.org/) C library
    + will also require a omnibus software definition but for ICU - also for which there is [some prior art in Habitat](https://github.com/habitat-sh/core-plans/blob/d611be73c0314d72eb4cc826c7a0fc54c64b81d9/icu/plan.sh)
+ in Ruby using file size and a "probably binary" threshold value

This PR represents an attempt at the first option which adheres the closest to how it seems most people think to solve the problem (i.e. "just use `file`") without having to shell out. The current state is the logical meat of the detection and worth reviewing for sanity before the standard boilerplate of a fieri metric worker, supermarket metric result, and endpoint get created.

One thing I'm thinking is that it might be worth _both_ the file type check and a file size check to maybe catch uu- or base64-encoding [shenanigans](https://raw.githubusercontent.com/poise/python/711e4d98c25643be2df5afaf3cfc2e811eaee18f/files/default/get-pip.py).

@thommay You shared an opinion about this earlier this week, so GOOD NEWS, you get to review as well!

thommay:
yeah, i think the file size check is useful, for the encoding reasons you mention - but also, because I fail to see why anything on the public supermarket should really contain a file over a meg or so.

nellshamrell:
I like the file magic approach you have taken thus far, I'm hesitant to go down the route of needing to add something to omnibus unless we truly have to, so that approach works for me.

robbkidd:
>  I fail to see why anything on the public supermarket should really contain a file over a meg or so.

@thommay So maybe a addendum in the feedback like:

```
Failure: To pass this metric, you should remove any binary files from the produced tarball.

Binaries found:
+ apache2/files/default/some-sort-of-binary.tar.gz
+ apache2/files/default/base64-encoded-shenanigans.rb - size > 1MB which seems kinda unnecessary
```

nellshamrell:
Per offline conversation, I am fine with adding an omnibus software definition to the supermarket omnibus files. At first I thought this involved making a change to the upstream omnibus software, which I was hesitant to, but this is fine!

image

robbkidd:
Wunderbar! Yes, we can control the addition today within our omnibus definitions. At some future point, the `file` software definition can move out to [omnibus-software](https://github.com/chef/omnibus-software) if another project needs it.

PR #1575, opened 2017-03-27
add an exception handler for not finding a repo

robbkidd:
Sometimes metadata declares a repo that doesn't exist anymore. Catch that and return an empty list of tags to fail the version tag check.


iennae:
👍

PR #1580, opened 2017-03-30
Use Foodcritic 10.2

tas50:


robbkidd:
@nathenharvey Because Foodcritic 10.2 now has a rule for asserting the presence of a `supports`, I propose that we depend on Foodcritic for [QM006](https://github.com/chef-cookbooks/cookbook-quality-metrics/blob/master/quality-metrics/qm-006-supports.md) and for this PR to also _remove_ the standalone SupportedPlatformsWorker.

PR #1583, opened 2017-04-04
add some more guards within metrics involving source repos

robbkidd:
use `present?` to fail on empty strings

Sometimes the `source_url` for a cookbook is an empty string, not a null. The `source_repo` metric will then return the empty string. So check for presence instead of nilness.

TODO
- [ ] figure out the `TypeError: no implicit conversion of String into Integer` in [`VersionTagWorker#tag_names`](https://github.com/chef/supermarket/blob/master/src/fieri/app/models/version_tag_worker.rb#L42)


robbkidd:
Regarding `TypeError: no implicit conversion of String into Integer`, it comes from the new GitHub API redirect behavior.

```
> octoclient.tags('rithis-chef/cookbook-youtrack')
=> {:message=>"Moved Permanently",
 :url=>"https://api.github.com/repositories/8707839/tags",
 :documentation_url=>"https://developer.github.com/v3/#http-redirects"}
```

We're currently using Octokit v3.8.0. v4 added automatic redirect following for the new behavior, so I'll work on upgrading the client tomorrow.

PR #1584, opened 2017-04-04
don't treat empty files as binary

robbkidd:
Zero length files aren't binary. They don't make much sense being
included in a cookbook, but they shouldn't cause the no binaries check
to fail.


src/fieri/app/models/cookbook_artifact.rb
```diff
@@ -113,7 +113,7 @@ def binary?(filepath)
       # need to check for application/xml explicitly because only versions of
       # libmagic < 5.26 are available on Ubuntu 14.04 and 16.04 which are our
       # CI (Travis and Automate) node platforms
-      !(magic.file(filepath) =~ %r{^(text\/|application\/xml)})
+      !(magic.file(filepath) =~ %r{^(text\/|inode\/x-empty|application\/xml)})
```
robbkidd: I think I'll want to refactor this bit the next time something gets added to the regex to declare yet another thing "not binary."

PR #1577, opened 2017-03-29
Accept either SPDX standard licenses or DK generated licenses

tas50:
Signed-off-by: Tim Smith <tsmith@chef.io>

PR #1586, opened 2017-04-05
Add additional acceptable license strings

### LINKED ISSUE #1585: Correct list of licneses
`knife cookbook create bar -I apachev2` creates a cookbook with the following license:

`license          'Apache 2.0'`

The current quality metric allows for the SPDX version (`Apache-2.0`) and for the short license name passed as an argument to knife (`apachev2`) but does not currently allow for the license string generated by knife (`Apache 2.0`).

For now the quality metric should allow all both the arguments to knife (`apachev2, gplv2, gplv3, mit`) and the values that are generated based on those arguments:

* `Apache 2.0`
* `GNU Public License 2.0`
* `GNU Public License 3.0`
* `MIT`

nathenharvey:
Fixes #1585

Signed-off-by: Nathen Harvey <nharvey@chef.io>

### DIFF
--- a/src/fieri/app/models/license_worker.rb
+++ b/src/fieri/app/models/license_worker.rb
@@ -6,10 +6,10 @@ class LicenseWorker
 
   # Acceptable licenses were determined and documented in
   # https://github.com/chef-cookbooks/cookbook-quality-metrics/blob/master/quality-metrics/qm-003-license.md
-  ACCEPTABLE_LICENSES = ['Apache-2.0', 'apachev2',
+  ACCEPTABLE_LICENSES = ['Apache-2.0', 'apachev2', 'Apache 2.0',
                          'MIT', 'mit',
-                         'GPL-2.0', 'gplv2',
-                         'GPL-3.0', 'gplv3'].freeze
+                         'GPL-2.0', 'gplv2', 'GNU Public License 2.0',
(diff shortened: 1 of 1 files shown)


nathenharvey:
@robbkidd I realize I might be late to the discussion here.

robbkidd:
@tas50 and I went back and forth on this and opted for SPDX in #1577. I think we both missed that the long versions come from knife, too. So, 👍 .

Do you think there is follow-on work that we should open a PR against the generator to use the SPDX license format? And then eventually circle back to this metric to reduce the list?

tas50:
I didn't realize that knife was where those strings were coming from. I'm going to make sure that gets cleaned up in Chef 13.

robbkidd:
Should we not also update the license metric definition to say these _three_ forms of strings are acceptable? Or leave the definition as is and remove the old style after Chef 13?

tas50:
I'd prefer we actually get people on the right track here, but we need to get our tools generating the acceptable license string. We're all over the place right now in the worst possible way.

robbkidd:
Ya. My current thinking is to allow these three styles until we've got the tools (knife, chef generate, foodcritic) all aligned and encouraging the SPDX format. Then we can _remove_ this code from Supermarket as the metric will be met by a new foodcritic rule in the supermarket category for open source licensing.

PR #1608, opened 2017-05-17
remove CLA/Curry signature tracking component

robbkidd:
* obviously removed a loooooot of code
deleting code

* `app/views/resources_collaborators` :: moved and renamed some helper methods used here from contributors_helper to collaborators_helper that were being used to render links for removing collaborators
* `spec/controllers/accounts_controller_spec` :: changed the path stored in stored_location for testing redirects to a path that exists now that there is no such thing as an icla_signature
* added `Curry::CommitAuthor` and `Curry::PullRequestCommitAuthor` directly to a migration that requires those classes and their logic to be present for the migration to succeed
* added a location matcher for `/curry/*` to web app's nginx config and to the app's routes to return `410 Gone` to provide a more useful error message to any GitHub repos that may have the webhook still enabled


src/supermarket/config/routes.rb
```diff
@@ -73,48 +73,13 @@
...
-  end
-
-  namespace :curry, constraints: proc { ROLLOUT.active?(:cla) && ROLLOUT.active?(:github) } do
-    resources :repositories, only: [:index, :create, :destroy] do
-      member do
-        post :evaluate
-      end
-    end
-
-    resources :pull_request_updates, only: [:create]
```
robbkidd: I could maybe see keeping this route—the former GitHub webhook handler—to return a 4xx something "Nothing to be found here anymore" error.

robbkidd: @thom I propose we change route from a removal to return `410: Gone` and an appropriate message. I see there is at least one project on GitHub still firing back to Curry.

src/supermarket/config/routes.rb
```diff
@@ -73,47 +73,14 @@
...
-  end
+  post 'curry/pull_request_updates', to: proc { [410, {}, ['Curry/CLA feature has been removed.']] }
```
robbkidd: This is the only route I see as needing something returned about its removal.

omnibus/cookbooks/omnibus-supermarket/templates/default/rails.nginx.conf.erb
```diff
@@ -85,6 +85,10 @@ server {
     break;
   }
 
+  location ~ ^/curry/ {
+    return 410;
```
robbkidd: Fast-fail krufty curry webhooks without making the Rails app think about it. While this kinda obviates the need for the Rails app to also handle this route, I've left an entry in routes.rb with the same behavior Just In Case™.

robbkidd:
While Travis is stuck, Unit and Lint passed in the pipeline.

PR #1615, opened 2017-06-09
fix lib class loading & const setting in production

robbkidd:
## The Trouble with Libraries

Rails 5 no longer auto-loads from the `autoload_paths` in production.

Options were to:
* re-enable that behavior
* add them to `eager_load_paths` insteads
* move to a directory under `app/` which is eager loaded by default

Opted for the latter to follow what seems to be the new convention.

## ROLLOUT Rolled No More

Removed the fancy `Object.const_missing` wrapper from around setting up Rollout.

This was causing some grief in production (but not dev/test). Removing the fancy wrapper and just setting the ROLLOUT constant directly in this initializer because initializers only run once.


robbkidd:
🤔  I see that the `Object.const_missing` trick was to allow some rake tasks without requiring an operational Redis, e.g. asset compilation. (ref #846)

I'll research more on why the trick isn't working in production environment.

PR #1616, opened 2017-06-09
wrap Rollout in a strangler class

robbkidd:
In an effort to own the feature flipping namespace and play well with other classes. By these means, we'll:

+ avoid humans needing to remember feature flippers are implemented in Rollout
+ avoid humans needing to remember to prepend :: to a constant to look up a feature flag's state
+ get errors in Dev and Test when the Feature class isn't resolved for some reason instead of ROLLOUT const resolving everywhere but production
+ make clear what the "API" is for how feature flags behave (which is why this is not implemented as a magical delegation through method_missing)

[Strangler Application](https://www.martinfowler.com/bliki/StranglerApplication.html)


robbkidd:
@thommay What I showed you this morning.

@nellshamrell If you've got time, this is the Rollout Strangler we spoke about last week.

PR #1618, opened 2017-06-14
upgrade foodcritic to 11.2

robbkidd:
Because.

### DIFF
--- a/src/fieri/Gemfile.lock
+++ b/src/fieri/Gemfile.lock
@@ -81,7 +81,7 @@ GEM
     erubis (2.7.0)
     faraday (0.12.1)
       multipart-post (>= 1.2, < 3)
-    foodcritic (11.1.0)
+    foodcritic (11.2.0)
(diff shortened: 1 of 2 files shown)


robbkidd:
@tas50 Any reason to _not_ move to foodcritic 11.2? I'm distracted and so I'm unsure of the ramifications listed in the release notes.

tas50:
We'll need to update the tags we use to any since 11.2 disabled the opensource tag.

robbkidd:
So we add `opensource` to the existing:

```
FIERI_FOODCRITIC_TAGS="metadata,correctness,supermarket,portability,deprecated ~FC031"
```

... ?

tas50:
Oh yes I forgot we have a curated list. We're good then

tas50:
If we want to check for the opensource license yes

robbkidd:
The one rule tagged with `opensource` at the moment is also tagged with `supermarket`, so we've opted to leave as is for the moment.

PR #1623, opened 2017-06-19
use Forwardable for all the feature flag delegation

robbkidd:
Be clear about:

1. Cannot instantiate a Feature object (module instead of class)
2. List of methods delegated to the "feature flag adapter" which happens to be Rollout in our implementation.

Thanks to @cupakromer and @keithrbennett for helpful suggestions.

### DIFF
--- a/src/supermarket/app/lib/feature.rb
+++ b/src/supermarket/app/lib/feature.rb
@@ -1,21 +1,16 @@
 require 'redis'
 
-class Feature
+module Feature
+  extend Forwardable
+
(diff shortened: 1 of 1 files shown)


jjasghar:
Walking through the code, it the less verbose makes more sustainable sense. The before merging though I suggest a comment or two explaining the `module_function` and `def_delegators` and the logic in using them. Also, adding a passing and expected fail test around this can with the logic explanation.

robbkidd:
@sigje @jjasghar Added more commentary to both the commit message and a code comment. 👀

PR #1629, opened 2017-07-06
Add AURD Terraform + Cookbook

chef-delivery:
This Pull Request was opened by Chef Automate user Tom Duffield

* Add terraform to stand up AURD
* Add supermarket-deploy cookbook to configure AURD
* Update travis.yml to execute terraform lint tests

Signed-off-by: Tom Duffield <tom@chef.io>

Ready to merge? View this change in Chef Automate and click the Approve button: https://automate.chef.co/e/chef/#/organizations/products/projects/supermarket/changes/8650b71e-61c7-4cf6-a039-61812db76fc4

### DIFF
--- a/.delivery/build_cookbook/libraries/helpers.rb
+++ b/.delivery/build_cookbook/libraries/helpers.rb
@@ -64,6 +64,18 @@ module BuildCookbook
     end
 
     #########################################################################
+    # Deploy Helpers
+    #########################################################################
+
+    # Determine which recipe to use for the given scenario
+    def recipe_for(infra_node)
+      package_type = infra_node.attribute?('scenario_package_type') ? infra_node['scenario_package_type'] : 'omnibus'
+      topology = infra_node.attribute?('scenario_topology') ? infra_node['scenario_topology'] : 'standalone'
+
+      "#{package_type}_#{topology}"
+    end
+
+    #########################################################################
     # Inspec
     #########################################################################
     # Coming soon!
--- a/.delivery/build_cookbook/metadata.rb
+++ b/.delivery/build_cookbook/metadata.rb
@@ -2,7 +2,7 @@ name 'build_cookbook'
 maintainer 'The Authors'
 maintainer_email 'you@example.com'
 license 'all_rights'
-version '0.2.1'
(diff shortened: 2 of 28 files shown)


robbkidd:
I think I understand what's going on in this change. I'm mostly 👍, though I'm curious if most of the things added could be put in a new `deploy/` directory at the root of the project. In an effort to keep the root of repo simple and put related things together where they can be documented and changed together. I admit this could be some old fashioned fastidiousness, but from a parallel effort in packaging with Habitat scaffolding-ruby, the root of the repo might get super busy with the rails app moving back there.

So maybe like ...?

```
+ - .delivery
+ - .expeditor
+ - deploy
    + - cookbooks
    + - scripts
    + - terraform
+ - omnibus
+ - src
```

tduffield:
So we can have that conversation, but "deploy" isn't really a good "container" for those files as they aren't strictly related to deployment. The scripts are used for travis, the cookbooks are used for deployment, and terraform is used for provisioning. It's also the same folder structure that we are using in other projects so on-call staff will be able to quickly find the files they are used to.

schisamo:
Yeah I'm pretty 👎  on us using a different structure _at this moment in time_. If we want to refactor things in the future it should be applied across all projects. It might be easier to just make the Habitat scaffolding work with Rails apps not at the root of a repo also. That feels like a general feature many Habitat end users would want...

robbkidd:
> It might be easier to just make the Habitat scaffolding work with Rails apps not at the root of a repo 

That change is crunching away in my hind brain waiting on foreground cycles. I would prefer that.

Also :lol: "easier"

robbkidd:
> I'm pretty 👎 on us using a different structure at this moment in time. If we want to refactor things in the future it should be applied across all projects.

Alrightie. I get that.

rhass:
@robbkidd I think all you might need to do to get the ruby scaffolding to work with Rails apps which are not in the root path of the repo is to set `SRC_PATH` in your `plan.sh` -- it should respect the value correctly.

robbkidd:
@rhass I've got some other gotchas I hope to have PRs or Issues opened for soon.

PR #1648, opened 2017-08-05
Use Foodcritic 11.3

tas50:


robbkidd:
I pushed an amended commit with the Gemfile.lock updated under `src/supermarket`, too. Gotta update both. I wonder if we ought to put a `foodcritic >= <desired version>` in fieri's gemspec so that we specify the version when we want to upgrade.

robbkidd:
Closing to resubmit via Automate.

PR #1665, opened 2017-08-31
Fix the activity feed, give credit where it is due

### LINKED ISSUE #1509: Supermarket shows the wrong user icons in the cookbook feed
In the cookbook release feed when a contributor releases a cookbook we're displaying the icon of the owner and not the user that shipped it.

screen shot 2016-12-30 at 10 29 18 pm

In this example I'm being given the icon for chef-cookbooks and sous-chefs depending on what cookbook I'm shipping.

This screenshot is from https://supermarket.chef.io/dashboard

davidalpert:
As described in Issue #1509 the fix for Issue #965 (PR #1293)
[shows the cookbook owner's avatar while displaying the name
of the person publishing the version](https://github.com/chef/supermarket/pull/1293/files#diff-0727a69a91c2b7e8e0dcf69328b22e0bR21).

This commit updates the message as per discussion on #1509:

> 👩  Chef: rsyslog v5.1.0 released by Timothy Smith 👱 

It also adjusts the feed to read the same way, though
without the leading leading and trailing gravatar

This strikes a balance in the feed between owner/maintainer
and publisher, so that the owner/maintainer's name still
leads the atom entry's subject, but the atom entry is
authored/published by the same person who published
the cookbook version.

Signed-off-by: David Alpert <david@spinthemoose.com>

---

Fixes #1509


davidalpert:
looks like some cleanup to do with those failing tests. will look into it tomorrow.

src/supermarket/app/views/pages/dashboard.html.erb
```diff
@@ -18,12 +18,14 @@
         <% @followed_cookbook_activity.each do |cookbook_version| %>
           <li>
             <div>
-              <%= link_to gravatar_for(cookbook_version.owner), cookbook_version.published_by %>
+              <%= link_to gravatar_for(cookbook_version.owner), cookbook_version.owner %>
               <%= t('cookbook.activity',
-                    maintainer: link_to(cookbook_version.published_by.name, cookbook_version.published_by),
+                    maintainer: link_to(cookbook_version.maintainer, cookbook_version.owner),
```
robbkidd: Despite the parameter name to the I18n renderer being `maintainer`, here you'll want `cookbook_version.owner.name` instead of `cookbook_version.maintainer`.

src/supermarket/app/views/users/followed_cookbook_activity.atom.builder
```diff
@@ -4,8 +4,8 @@ atom_feed language: 'en-US' do |feed|
 
   @followed_cookbook_activity.each do |cookbook_version|
     feed.entry cookbook_version, url: cookbook_version_url(cookbook_version.cookbook, cookbook_version.version) do |entry|
-      entry.title t('cookbook.activity',
-                    maintainer: cookbook_version.published_by.name,
+      entry.title t('cookbook.activity_feed',
+                    maintainer: cookbook_version.maintainer,
```
robbkidd: `cookbook_version.owner.name` here, too.

src/supermarket/config/locales/en.yml
```diff
@@ -38,7 +38,8 @@ en:
     undeprecated: "%{cookbook} is no longer deprecated."
     featured: "%{cookbook} is now %{state}."
     deprecate_with_deprecated_failure: "This cookbook cannot be deprecated in favor of a deprecated cookbook."
-    activity: "%{maintainer} released v%{version} of the %{cookbook} cookbook."
+    activity: "%{maintainer}: %{cookbook} v%{version} released by"
```
robbkidd: Ugh. I'd forgotten that the text was handled by an I18n entry. "released by ..." and then the rest of the statement finished back in the view makes my brain itch.

🤔 

How awful would ...

👩‍🍳 Chef: rsyslog v5.1.0 released by Timothy Smith 👱 

... look? I think I'm OK with that. Curious what you think. That would let us keep a single I18n `activity` statement—it would need to take a new parameter like `publisher`—and only do the gravataration in the view.

davidalpert: that makes sense to me; the awkwardness made me itch too. let me spin that up and we'll see how it looks.

robbkidd:
An example from my dev environment:

screen shot

robbkidd:
Now with less margin!

screen shot with no margin

PR #1666, opened 2017-09-01
update the contributor guide

davidalpert:
add some detail about how to:

- run postgres and redis through the `docker-compose.yml` file;
- resolve some potential issues installing the gem bundle;
- run the specs locally;
- run rubocop locally;

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,15 +45,88 @@ Bad:  Added some feature
 Bad:  Adding some feature
 Good: Add some feature
 ```
+### Running Locally
+
+##### Known issues with `bundle install`
+
+- the `ruby-filemagic` gem fails to build natively:
+  ```
+  Gem::Ext::BuildError: ERROR: Failed to build gem native extension.
+
+      current directory:  /path/to/your/gems/ruby-filemagic-0.7.1/ext/filemagic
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -45,15 +45,88 @@ Bad:  Added some feature
...
+  No pg_config... trying anyway. If building fails, please try again with
+   --with-pg-config=/path/to/pg_config
+  checking for libpq-fe.h... no
+  Can't find the 'libpq-fe.h header
+  *** extconf.rb failed ***
+  ```
+
+  For Linux: `sudo apt-get install libpq-dev`
+
+  For Mac: `brew update` then `brew install prostgresql`
```
tas50: postgresql is spelled wrong here

CONTRIBUTING.md
```diff
@@ -45,15 +45,88 @@ Bad:  Added some feature
 Bad:  Adding some feature
 Good: Add some feature
 ```
+### Running Locally
```
tas50: You should probably mention that you need to be in the src directory for supermarket in order to do the bundle install

tas50: It might also be worth mentioning that you should install the version of Ruby that Supermarket uses

PR #1681, opened 2017-10-12
remove some interpolated default attributes for sensible ocid default

robbkidd:
This is the start of removing interpolated values from the omnibus' cookbook attribute defaults. Interpolating the default chef_server_url into chef_identity_url in the defaults caused no end of trouble with onprem installations. The [application URL helpers](https://github.com/chef/supermarket/blob/c6970b5aeb95746c24386e7c8cb88000ea5c8790/src/supermarket/app/helpers/custom_url_helper.rb) already perform run-time interpolation with the chef_server_url when no env var is present. So we'll leave these attributes unset which will omit them from the list of things that get exported as env vars.

Administrators still have the option to set a value for these attributes which will set the associated env var and will override in the Rails app.

### DIFF
--- a/omnibus/cookbooks/omnibus-supermarket/attributes/default.rb
+++ b/omnibus/cookbooks/omnibus-supermarket/attributes/default.rb
@@ -340,19 +340,27 @@ default['supermarket']['api_item_limit'] = 100
 # ### Chef URL Settings
 #
 # URLs for various links used within Supermarket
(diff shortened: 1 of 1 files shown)


robbkidd:
@sigje @nellshamrell 👀 

@larryebaum This change, once released, should address the OCID stuff needed for onprem Supermarkets. After this change, you shouldn't need to override `["supermarket_omnibus"]["config"]["chef_identity_url"]` any more. What the docs say should be sufficient once updated to talk about setting SSL and the FQDN if using a CNAME.

PR #1698, opened 2017-11-23
Convert S3_URLS_EXPIRE to a number on use

KierranM:
I'm attempting to set up a private supermarket that is backed with S3 and uses presigned urls for access when downloading cookbooks.

While doing this I found that setting the `S3_URLS_EXPIRE` environment variable causes the cookbook download to fail with a 500 error. The error in the log states: `ArgumentError (comparison of String with 604800 failed)`. After some digging I found that the issue is caused by the `S3_URLS_EXPIRE` environment variable is always a string, and gets passed down to the aws-sdk via paperclip and hits a comparison that ensures the given expiry time is less than one week (604800 seconds).

This PR fixes this by always converting the string to a number when asking for the expiry_url.

I haven't been able to set up a proper dev environment for this, so I'm unsure if the tests actually run.


KierranM:
Ah damn, I'm on a different laptop so can only edit through GitHub, I'll squash and signoff the commits when I'm back home

robbkidd:
Oooo, nice! Good catch, @KierranM. Puttin' this in the review queue.

PR #1702, opened 2017-12-07
upgrade omnibus for new openssl

robbkidd:


robbkidd:
While this branch tests OK, it currently descends from an omnibus configuration that does not build. See #1703, which will likely need to merge first to fix the build.

PR #1703, opened 2017-12-07
fix omnibus build, pin rubygems to 2.6.x

robbkidd:
RubyGems 2.7.0-.3 seems to break the omnibus build with:

```
ERROR:  Error installing bundler:
"bundle" from bundler conflicts with /opt/supermarket/embedded/bin/bundle
```

Pinning to latest 2.6.x to fix the pipeline for release.

PEOPLE OF THE FUTURE, this pin should get revisited.


robbkidd:
Tests pass here and the omnibus build was successful. ✅

PR #1704, opened 2017-12-08
reorganize the fieri subcomponent to make supermarket more container friendly

robbkidd:
Move the fieri engine to a directory under the supermarket Rails root.

This is to admit that the fieri engine "service" is really an engine that runs within the running supermarket app itself. Until we want to run it separately, it will be better to place the gem source somewhere under the supermarket application source to make possible packaging the app as a single thing without path shenanigans.

Also, stop using the repo's README as a sample README for seed data because the repo's README isn't available within a containerized or habiterized Supermarket app.


robbkidd:
Testing an omnibus build in the pipeline.

robbkidd:
Omnibus build succeeded. Tested uploading a cookbook to a test instance of that build. With fieri relocated, it still processed the cookbook uploaded and produced quality metric results.

.travis.yml
```diff
@@ -9,7 +9,8 @@ cache:
   bundler: true
   directories:
     - $HOME/tools
-    - "$TRAVIS_BUILD_DIR/src/$COMPONENT/vendor/bundle"
+    - src/supermarket/vendor/bundle
+    - src/supermarket/engines/fieri/vendor/bundle
```
robbkidd: Opting to be explicit about caching the bundled gems in the two Ruby code bases.

.travis.yml
```diff
@@ -44,15 +45,16 @@ matrix:
...
       before_install:
-        - cd $TRAVIS_BUILD_DIR/src/$COMPONENT && bundle config build.nokogiri --use-system-libraries
+        - cd src/supermarket
+        - bundle config build.nokogiri --use-system-libraries
       before_script:
-        - cd $TRAVIS_BUILD_DIR/src/$COMPONENT && bundle exec rake $COMPONENT:travis:before_script
+        - bundle exec rake supermarket:travis:before_script
       script:
-        - cd $TRAVIS_BUILD_DIR/src/$COMPONENT && bundle exec rake $COMPONENT:travis:script
+        - bundle exec rake supermarket:travis:script
```
robbkidd: The `cd src/supermarket` in `before_install` was sufficient for getting into the working directory needed for testing. This matches the style of the other test jobs added to Travis for the pipeline.

.travis.yml
```diff
@@ -64,11 +66,12 @@ matrix:
...
       before_install:
-        - cd $TRAVIS_BUILD_DIR/src/$COMPONENT && bundle config build.nokogiri --use-system-libraries
+        - cd src/supermarket/engines/fieri
+        - bundle config build.nokogiri --use-system-libraries
       before_script:
-        - cd $TRAVIS_BUILD_DIR/src/$COMPONENT && bundle exec rake $COMPONENT:travis:before_script
+        - bundle exec rake fieri:travis:before_script
       script:
-        - cd $TRAVIS_BUILD_DIR/src/$COMPONENT && bundle exec rake $COMPONENT:travis:script
+        - bundle exec rake fieri:travis:script
```
robbkidd: The `cd src/supermarket/engines/fieri` in `before_install` was sufficient for getting into the working directory needed for testing. This matches the style of the other test jobs added to Travis for the pipeline.

omnibus/config/software/fieri.rb
```diff
@@ -1,22 +0,0 @@
...
-# See the License for the specific language governing permissions and
-# limitations under the License.
-#
-
-name "fieri"
-license :project_license
-
-source path: File.expand_path('../../../../src/fieri', project.filepath)
-
-dependency "file"
```
robbkidd: Because fieri didn't exist within supermarket's Rails root directory structure, these were the shenanigans necessary to get the fieri source located in a spot that the supermarket Gemfile expected (`../fieri`) for the omnibus build.

omnibus/config/software/supermarket.rb
```diff
@@ -30,7 +30,7 @@
 dependency "ruby"
 dependency "runit"
 dependency "logrotate"
-dependency "fieri"
+dependency "file"
```
robbkidd: With this change fieri lives as a gem whose files are located within the supermarket source tree. Fieri uses `ruby-filemagic` which is a wrapper with C extensions for `libmagic` in `file`, so supermarket gets the dependency.

src/supermarket/Gemfile
```diff
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 # Ruby version isn't specified here anymore, as it's pinned in Omnibus
 # https://github.com/chef/omnibus-supermarket/blob/master/config/projects/supermarket.rb
 
-gem 'fieri', path: '../fieri'
+gem 'fieri', path: 'engines/fieri'
```
robbkidd: The tip of the ice berg. With fieri's files now located under the supermarket application root, the app root can now be the top-most directory included in a habitat package or container. Build and runtime no longer have to reach outside the app root tree.

src/supermarket/db/seeds.rb
```diff
@@ -88,7 +88,7 @@
         description: Faker::Lorem.sentences(1).first,
         license: 'MIT',
         tarball: File.open('spec/support/cookbook_fixtures/redis-test-v1.tgz'),
-        readme: File.read('../../README.md'),
+        readme: '# This is a README',
```
robbkidd: Removing another file reference that is reaching outside of the supermarket app root tree. For a long time the project's README has been used as a fixture README for development environment seeding of test cookbooks. This stops that. If we opt to have more complicated test READMEs, they should probably live in the test fixtures directory.

pwelch:
Looks good to me 👍

.travis.yml
robbkidd: Coming back to this, I think we should add `$TRAVIS_BUILD_DIR` back in if the goal is explicitness.

robbkidd:
I think this is ready. :shipit:

PR #1710, opened 2018-01-03
pin omnibus build's berkshelf version to latest stable

robbkidd:
There is a major version bump for berkshelf in progress on its master branch. The software definition for berkshelf rides master, so we're pinning to latest released version of berkshelf 6.


robbkidd:
This got lumped into #1716 which will be shipping shortly.

PR #1715, opened 2018-01-18
Fix version validation against chef's required format

### LINKED ISSUE #1714: Validate cookbook version format during upload
It is possible in Supermarket versions < 3.1.53 to upload a cookbook with a version number that does not conform to [Chef's cookbook version format of either X.Y.Z or X.Y](https://docs.chef.io/cookbook_versions.html#constraints). This was discovered recently when [a community cookbook was uploaded with a semantic version that included a label](https://github.com/Cognifide/aet-cookbook/issues/12).

### with `knife supermarket` ✅ 

An upload will correctly fail with the following error:

```
ERROR: Chef::Exceptions::InvalidCookbookVersion: '1.2.3-rc4' does not match 'x.y.z' or 'x.y'
```

In the process of preparing an upload, `knife supermarket` instantiates a `Chef::Version` which [performs some validation and throws an exception for non-compliant versions](https://github.com/chef/chef/blob/b12f3cc1e899006ac01bfef9348b89e19ab5027c/lib/chef/version_class.rb#L62-L72).

### with `stove` ❌ 

A cookbook with a version that is a valid semantic version (e.g. `1.2.3-rc4`, note the additional label allowed by SemVer) uploaded with `stove` will succeed. The UI will no longer display the pages for this cookbook or its previous, valid versions because of a `digits and dots only` matcher for routing. This renders the cookbook unmanageable through the web UI and the problematic version unmanageable through the API. To fix, the cookbook version must be destroyed from the Supermarket Rails console.

### with `knife supermarket` (unchanged)

An upload correctly fails with the following error:

```
ERROR: Chef::Exceptions::InvalidCookbookVersion: '1.2.3-rc4' does not match 'x.y.z' or 'x.y'
```

### with `stove`

An upload correctly fails with something like the following error:

```
E: Stove experienced an error!
E: ChefAPI::Error::HTTPBadRequest
E: The Chef Server did not understand the request because it was malformed.

    {"error_code"=>"INVALID_DATA", "error_messages"=>["Version is not a valid Chef version. Chef::Exceptions::InvalidCookbookVersion: '1.2.3-rc4' does not match 'x.y.z' or 'x.y'"]}
```

## What's up with that?

Two reasons a cookbook can be uploaded with a version that does not match Chef's version format:

* PRIMARY: Supermarket accepts the upload because as late as version 3.1.53, [cookbook versions were being validated against SemVer](https://github.com/chef/supermarket/blob/7d6e7e3718c54a53189f38b2813724f334169c68/src/supermarket/app/models/cookbook_version.rb#L101-L111) not against Chef's version format definition which is more strict than SemVer (i.e. no labels).
* SECONDARY: `stove` does not perform validation on version format conforming to Chef requirements.

robbkidd:
[A version for a cookbook must be either X.Y.Z or X.Y](https://docs.chef.io/cookbook_versions.html#constraints). Up until this change, the validation for a CookbookVersion's version was against SemVer rules. SemVer allows for release labels or build info to appear after the version number itself. Chef does not!

Here, we add a new ChefVersionValidator that reuses the existing version validation logic of the Chef gem in ::Chef::Version. Supermarket is already relying on the Chef gem to validate version constraints set for cookbook dependencies and supported platform.

Fixes #1714 to return an error to any client if a cookbook is uploaded with an invalid version.

### DIFF
--- a/src/supermarket/Gemfile
+++ b/src/supermarket/Gemfile
@@ -66,7 +66,6 @@ group :development do
   gem 'guard-rspec', require: false
   gem 'guard-rubocop', require: false
   gem 'license_finder'
-  gem 'pry-rails'
   gem 'spring'
   gem 'spring-commands-rspec'
 end
@@ -91,6 +90,7 @@ group :development, :test do
   gem 'byebug'
   gem 'launchy'
   gem 'mail_view'
+  gem 'pry-rails'
   gem 'rspec-rails'
   gem 'rubocop', '>= 0.23.0'
 
--- a/src/supermarket/Gemfile
+++ b/src/supermarket/Gemfile
@@ -43,7 +43,6 @@ gem 'rinku', require: 'rails_rinku'
 gem 'rollout'
 gem 'sass-globbing'
 gem 'sass-rails'
-gem 'semverse'
 gem 'sentry-raven', require: false
 gem 'sitemap_generator'
 gem 'sprockets'
--- a/src/supermarket/Gemfile.lock
+++ b/src/supermarket/Gemfile.lock
@@ -524,7 +524,6 @@ GEM
     sawyer (0.8.1)
       addressable (>= 2.3.5, < 2.6)
       faraday (~> 0.8, < 1.0)
-    semverse (2.0.0)
     sentry-raven (2.4.0)
       faraday (>= 0.7.6, < 1.0)
     serverspec (2.38.1)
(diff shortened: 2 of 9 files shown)


robbkidd:
@thommay Because Supermarket is already depending on the Chef gem for `Chef::VersionConstraint` validation logic, is it sensible to double-down on the gem and use it for `Chef::Version` validation and comparison logic (instead of `berkshelf/semverse`)? Or do you foresee this leading down a dark road of pain and misery?

lamont-granquist:
This looks correct.  Chef::Version is what we use to validate cookbook versions internally.

PR #1716, opened 2018-01-19
Fix ctl-reconfigure failing in air-gapped environments

### LINKED ISSUE #1712: Supermarket reconfigure fails in airgapped environment, ignores rubygems_url
Attempted upgrade to supermarket 3.1.51 in airgapped environment because of #1705 using supermarket-omnibus-cookbook. Chef run to perform the upgrade failed during `supermarket-ctl reconfigure` step with error message (stdout and stack trace below) from bundler indicating that it couldn't reach https://rubygems,org/.

[stdout.txt](https://github.com/chef/supermarket/files/1604568/stdout.txt)
[stacktrace.txt](https://github.com/chef/supermarket/files/1604569/stacktrace.txt)

This system has rubygems_url set to our private gemserver in /etc/chef/client.rb, but supermarket seems to be ignoring that setting. I was able to get the reconfigure step to finish by changing the default gemserver on line 1055 of `/opt/supermarket/embedded/lib/ruby/gems/2.4.0/gems/chef-config-13.6.4/lib/chef-config/config.rb` from rubygems.org to our private gemserver - after that, the reconfigure ran just fine and the supermarket is back up and running now.

Am I missing something configuration-wise, or is this a regression since 3.1.42 (the last supermarket release I was successfully able to upgrade to)? I couldn't find a way to specify rubygems_url for supermarket specifically.

robbkidd:
**TL;DR:** ensure that the omnibus-internal cookbook for `reconfigure` does not depend on gems, directly or through a dependency cookbook.

Fixes #1712 

# What?

This moves the logic and dependencies of building an omnibus package in a development environment to a new cookbook. This separates building from the `omnibus-supermarket` cookbook internal to the omnibus package and what drives `supermarket-ctl reconfigure`.

# Why?

Up to this point, `omnibus-supermarket` depended on the `omnibus` cookbook. Its `build_package` recipe required the `omnibus` cookbook to setup a Test Kitchen VM with the kit necessary to build.

The `omnibus` cookbook requires the `pkg-config` gem. In Nov 2017, the installation of this gem moved from a converge-time `chef_gem` resource to a very-early-in-the-chef-run-all-the-time gem metadata entry. Gems declared in metadata get `bundle install`ed at the beginning of a chef
run regardless of whether the recipes/resources that use them are called.

This early-in-the-chef-run `bundle install` was causing `supermarket-ctl reconfigure` to fail in air-gapped environments where the host is unable to reach rubygems.org. The omnibus' chef-client run does not use a system's chef configuration (e.g. `/etc/chef/client.rb`) and it is prohibitively annoying to reach into the omnibus-generated `solo.rb` chef configuration to point it at an alternative gem server.


robbkidd:
Any thoughts on how to catch "this cookbook or one of its dependencies depends on a gem" in an automated test? The `Berksfile.lock` for `omnibus-supermarket` is intentionally not checked-in so as to track latest. But if one of the cookbooks it depends on adds a metadata gem dependency, we might silently reintroduce this bug.

omnibus/.kitchen.yml
```diff
@@ -9,26 +9,15 @@ driver:
 
 provisioner:
   name: chef_zero
-  require_chef_omnibus: 12.13.37
+  require_chef_omnibus: 13.6.4
```
tas50: Beautiful

omnibus/.kitchen.yml
```diff
@@ -9,26 +9,15 @@ driver:
...
 
 platforms:
-  - name: ubuntu-12.04
-    run_list: apt::default
+  - name: ubuntu-14.04
   - name: centos-5.11
-    run_list: yum-epel::default
   - name: centos-6.8
```
tas50: You should make this centos-6 and centos-7 which always point to the latest point release.

omnibus/cookbooks/omnibus-supermarket/.kitchen.yml
```diff
@@ -12,7 +12,6 @@ provisioner:
     ssl_verify_mode: verify_peer
 
 platforms:
-  - name: ubuntu-12.04
   - name: ubuntu-14.04
   - name: ubuntu-16.04
   - name: centos-5.11
```
tas50: We still doing RHEL 5?

omnibus/.kitchen.yml
robbkidd: YES I SHOULD

omnibus/cookbooks/omnibus-supermarket/.kitchen.yml
robbkidd: The answer was "yes" last time I asked, but maybe it's time to ask again. RHEL5 isn't on the [Platforms and Supported Versions list](https://docs.chef.io/platforms.html) any more, so maybe not.

tas50: We killed it Jan 1st

robbkidd:
An idea for testing "this cookbook or one of its dependencies depends on a gem": finally convert the omnibus-supermarket test suite to inspec, and then use the inspec `json` resource to assert that none of the files found with `find /opt/supermarket/embedded/cookbooks/ -name metadata.json -print` have a non-empty `gems` key.

Experimenting.

robbkidd:
@thommay I'd appreciate a re-review of the changes I've added. Maybe walk the last two commits. One adds the failure to omnibus build if it has pulled in cookbooks that depend on gems. The other is the serverspec->inspec conversion and I'm not 100% on the naming of the test groups or the friendliness of the output.

PR #1719, opened 2018-01-22
disable remote fetching in Paperclip

robbkidd:
Upgrade to Paperclip 5.2+. There is [a CVE in Paperclip v5.1.0](https://nvd.nist.gov/vuln/detail/CVE-2017-0889).

***** Supermarket was not vulnerable to this CVE. *****

The `tarball` parameter given for an upload is validated by `CookbookUpload::Parameters` and `CookbookUpload::Archive` before a hand-off to ActiveRecord+Paperclip for processing as an attachment.

This update to the Paperclip initializer is intended to prevent future accidental support for remote fetching if the logic of `CookbookUpload` changes at some point and allows a URL string through to the persistence layer. Paperclip 5.2.0 disabled these remote adapters. Leaving the initializer disabling in case that behavior changes.

Added tests to catch regressions.


robbkidd:
Or I could upgrade to Paperclip 5.2.0 that was just released with these adapters disabled by default.

robbkidd:
In addition to disabling the adapters in the initializer, I've upgraded to the latest Paperclip version that _also_ does not enable these adapters by default. Added tests to catch if the adapters sneak back in some how.

PR #1720, opened 2018-01-25
Update terraform to 0.11.2

jeremiahsnapp:


.travis.yml
```diff
@@ -11,6 +11,10 @@ cache:
     - $HOME/tools
     - "$TRAVIS_BUILD_DIR/src/$COMPONENT/vendor/bundle"
 
+before_install:
+  - curl https://raw.githubusercontent.com/chef/ci-studio-common/master/install.sh | bash
+  - export PATH="$PATH:$HOME/ci-studio-common/bin:$HOME/tools/bin"
```
robbkidd: I _think_ currently that the only job in the matrix that requires this `ci-studio-common` kit is the terraform linter. Maybe move this `before_install` to that job's entry instead of having all jobs retrieve it?

tduffield: The Release Engineering team is moving to leverage the ci-studio-common codebase to standardize all our pipeline testing across all our projects. While the linting phase is the only thing that leverages it now, we hope to introduce other things in the future. Some of these may not require modifying the individual test units themselves. As such, I'd prefer to keep it as is, unless you have a strong objection. If you do, I'm happy to hop onto video chat with you to discuss more.

robbkidd: I've got two thoughts/reservations:

* the `curl | sh` adds ~40 seconds to a test run
* `before_install` doesn't merge so the jobs that declare custom ones replace the common one; if a job needs the CI kit _and_ something else, that job will have to repeat the CI kit install anyway
    * This works out OK for the two application-level test jobs so that they _don't_ install the CI common kit when they don't need it. Those jobs are already 7 and 4 minutes and I'm trying to keep them from getting longer.

PR #1726, opened 2018-02-20
upgrade nokogiri & libxml2

robbkidd:
Upgraded the gem references for nokogiri and upgraded omnibus/-software for an updated libxml2 "system library." Addresses CVE-2017-15412

Also, remove `node.set` (deprecated) from dev build cookbook. `node.set` is deprecated. Use `node.default`/`node.override` ...  OR just set them in the kitchen config!

### DIFF
--- a/omnibus/.kitchen.yml
+++ b/omnibus/.kitchen.yml
@@ -18,5 +18,12 @@ platforms:
 
 suites:
   - name: default
+    attributes:
+      omnibus:
+        build_user: vagrant
+        build_user_group: vagrant
+        build_user_home: /home/vagrant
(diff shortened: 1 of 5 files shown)


omnibus/.kitchen.yml
```diff
@@ -18,5 +18,12 @@ platforms:
 
 suites:
   - name: default
+    attributes:
+      omnibus:
+        build_user: vagrant
+        build_user_group: vagrant
+        build_user_home: /home/vagrant
+        build_dir: /home/vagrant/supermarket/omnibus
+        install_dir: /opt/supermarket
```
tas50: Does this cookbook only ever run via TK?

PR #1732, opened 2018-03-30
Add BSD-3-Clause as a valid open source license

nathenharvey:
Signed-off-by: Nathen Harvey <nharvey@chef.io>


iennae:
Should this match https://github.com/Foodcritic/foodcritic/blob/master/lib/foodcritic/rules/fc078.rb?

robbkidd:
This _is_ the change to add that license to the list the license quality metric uses. So 👍.

And there's a part of me that would like to remove this license specific quality metric entirely in favor of FC078 that is run as a part of the Foodcritic quality metric. The list there is already more comprehensive.

nathenharvey:
catching this sooner and maintaining one list feels like the right thing to me!  I only added this one b/c I noticed that all of the facebook cookbooks were failing the metric on the Supermarket. 

Feel free to close this and/or remove all the license metrics...or ask me to :)

iennae:
Yeah, it feels like it should be one list rather than maintaining in 2 different places.

nathenharvey:
See #1733 wherein I just ripped out the license worker.  Hopefully this passes and is a better path forward ;)

PR #1733, opened 2018-04-02
remove license quality metric (covered by foodcritic rule 78)

nathenharvey:
This is covered by Foodcritic.

[FC078](http://www.foodcritic.io/#FC078)

TODO:
- [x] Keep the QualityMetric record named License which is associated with past results.
- [x] Remove LicenseWorker from Fieri
- [x] Deprecate the license_evaluation API endpoint
- [x] Update foodcritic_evaluation API endpoint to remove past License metric results from a cookbook version after creating a new Foodcritic result


robbkidd:
Thoughts on what—if anything—to do with existing License metric results in a system? Let them continue to exist on cookbooks for which the License metric was run? Which will leave a QualityMetric type for which there is no associated worker, probably requiring some commentary on that class about its continued existence.

nathenharvey:
Great question!

Given that Foodcritic is now taking responsibility for this check, would it makes sense to simply rerun Foodcritic on the most recent versions of all cookbooks and then remove this metric?

iennae:
Would love to pair up to discuss this at some point next week if folks are available? I think maybe a discussion about how to deprecate this and having a process for not having to change old quality metrics would be useful. Or quality metrics should only be visible for latest? which doesn't seem as useful either.

iennae:
Pings @robbkidd and @nathenharvey :) I'm still up for some pairing if you are

robbkidd:
@iennae I can make some time this week, yep.

robbkidd:
I've noodled an option for keeping historic results of the License metric while cleaning it up on cookbook versions that have had the newer Foodcritic run with the license rule.

1. All of this PR as is so far to remove the `LicenseWorker` from Fieri.
2. Remove the license metric endpoint from Supermarket that is responsible for recording new results:
    - [the route](https://github.com/chef/supermarket/blob/3.1.64/src/supermarket/config/routes.rb#L38)
    - [the API endpoint](https://github.com/chef/supermarket/blob/3.1.64/src/supermarket/app/controllers/api/v1/quality_metrics_controller.rb#L98-L109) the route ..er.. routes to
3. Keep the `QualityMetric` record named `License` which is associated with past results.
4. Update [the Foodcritic result API endpoint](https://github.com/chef/supermarket/blob/3.1.64/src/supermarket/app/controllers/api/v1/quality_metrics_controller.rb#L20-L31) to remove an existing License result from a cookbook version after the Foodcritic result has been recorded
    - This could also get tucked into the controller's [`create_metric` method](https://github.com/chef/supermarket/blob/3.1.64/src/supermarket/app/controllers/api/v1/quality_metrics_controller.rb#L279-L291) that is responsible for recording a new result and removing previous results. Would mean an `if` in there checking to see if it is creating a `Foodcritic` metric and to do a deletion similar to the existing removal of previous results, just with `License` as the `quality_metric` type.

iennae:
Hey @robbkidd want to have another go at this pairing so we can finish it out?

robbkidd:
@iennae I had a moment to work on this today. Turned out to be pretty simple.

I think this is ready to review and ship.

Issue #1684, opened 2017-10-17
Missing <pubDate> in RSS Feeds


pschaumburg:
We're using Slack for communication. We have added the RSS Feed App to one of our main Chef-Update Channels, where we receive different informations and also from updated Cookbooks inside our private Chef Supermarket. 

I've written already to Slack Support and gained the information, that the RSS Feed is good, but there is missing a <pubDate> tag. So the RSS Feed works only once and after that, it stops working.

Best regards,
Patrick

robbkidd:
Despite the icon showing "rss", the feed produced by Supermarket is in Atom format. Atom calls for an `<updated>` element that serves the same purpose as RSS's `<pubDate>`.

To which of the few lists of cookbooks have you subscribed?

robbkidd:
I can confirm some wonkiness with the feed. I subscribed to [the Supermarket's recently_updated feed](https://supermarket.chef.io/cookbooks.atom?order=recently_updated) in our Slack and I see some but not all of the recent cookbook versions that have been released appear. I see something similar when subscribed to the feed with another RSS reader.

My theory about what's going on has to do with how RSS clients treat entry `published` and `updated` dates. The feed entries are `cookbooks` and the AtomFeedHelper in Rails is using the cookbook's `created_at` date for the feed entry's `published` date and the cookbook's `updated_at` date for the feed entry's `updated` date. A cookbook's `updated_at` is the time that anything about the cookbook has changed (e.g. new version released, owner/collaborator changes, deprecations, etc). But the feed clients are preferring to order on entry `published`instead of on entry `updated`.

robbkidd:
Having watched the subscription over the weekend, I am more confident in my theory that feed entries are only appearing in Slack when the cookbook is new to the Supermarket instance (thus a fresh and recent `published` date for the feed entry) and not appearing for releases of an existing cookbook (where the `published` date for the entry remains in the past and only the `updated` date is ..er.. updated).

robbkidd:
@pschaumburg Can you confirm for this issue what feed URL path you have subscribed to in your Slack? I suspect it is `<your-supermarket-host>/cookbooks.atom?order=recently_updated`, yes?

pschaumburg:
@robbkidd That's the correct path.
I've also changed some informations with the Slack support. They gave me this answer, which also confirms it:

> It looks like your feed is missing the 
[...]

PR #1740, opened 2018-05-27
🚧WIP 🚧a new route for a simple "recent releases" feed

robbkidd:
This is very WIP. DO NOT MERGE

Intended to address #1684, this will add a new URL endpoint for a simple, reverse chronological atom feed of cookbook _releases_. The current feed available (described in #1684) is based on a `cookbooks/` endpoint that is scoped to the parent cookbook record and not individual released versions. It's that parent scoping that led to published dates tracking the first time a cookbook was released as opposed to when a particular version is released.

# Use Cases

To address a few different use cases for wanting notifications for new cookbook releases: 

## Curated: "I only care about these N cookbooks" ✅ 

This exists already. A user can see a feed of releases of cookbooks they follow. Anyone can subscribe to any user's feed, so this could be the means by which a single account could be used to curate a collection of cookbooks for consumption by many interested parties.

`<super-url>/users/<a_user>/followed_cookbook_activity.atom` 

## Branded: "I only care about cookbooks owned by X" 🆕

**New!** This would be a feed of releases owned by `username` regardless of the collaborator who published the release. This will be useful for release announcements for cookbooks owned by prolific users or organizations (e.g. Sous Chefs). 

`<super-url>/cookbooks/versions.atom?owned_by=<username>`

## Fire Hose: "All the releases, please!" 🆕 

**New!** This would be a feed of _all_ cookbook releases in a Supermarket instance. Probably more useful in private Supermarkets.

`<super-url>/cookbooks/versions.atom`

# Proposed Implementation

A new URL endpoint: `https://supermarket.example.com/cookbooks/versions.atom`

Endpoint would take an optional `owned_by` parameter which must be the username of a user in the Supermarket.

Endpoint responds to `.atom` with a feed:

```
Title: "Cookbook Releases"
Updated: (datetime of the latest updated_at among the cookbook versions)
```

 And has entries that look similar to the followed cookbook activity feed, something like:

```
Title: awesome_sauce 1.2.3
Author: Hot Sauce Fan (ilovehotsauce)
Link: http://localhost:3000/cookbooks/awesome_sauce/versions/1.2.3
Date: (this cookbook version's created_at datetime)

Sauce Chefs: awesome_sauce 1.2.3 released by Hot Sauce Fan

Installs/Configures awesome_sauce

(content from the CHANGELOG if present)
```

# TODO

- [ ] settle on a sensible URL path
    - `/cookbooks/versions`? - given that a feed of cookbook releases maps to a collection of `CookbookVersions`, this would follow the Rail convention for such a collection
    - `/recent_releases`? - maybe a more user friendly feed path, doesn't follow convention
- [ ] testing!
- [ ] support parameters to get a filtered feed; if so, what filters?
    - [ ] filter by owner
    - [ ] pagination

### DIFF
--- a/src/supermarket/app/models/cookbook_version.rb
+++ b/src/supermarket/app/models/cookbook_version.rb
@@ -11,7 +11,8 @@ class CookbookVersion < ApplicationRecord
   has_many :metric_results, -> { includes :quality_metric }, inverse_of: :cookbook_version
 
   belongs_to :cookbook
-  belongs_to :user
+  belongs_to :user, -> { includes :chef_account }, inverse_of: :cookbook_versions
+  has_one :owner, -> { includes :chef_account }, through: :cookbook
 
   # Attachments
   # --------------------
--- a/src/supermarket/app/views/cookbook_versions/index.atom.builder
+++ b/src/supermarket/app/views/cookbook_versions/index.atom.builder
@@ -1,0 +1,21 @@
+atom_feed language: 'en-US' do |feed|
+  feed.title 'Cookbook Releases'
+  feed.updated safe_updated_at(@cookbook_versions)
+
+  @cookbook_versions.each do |release|
+    feed.entry release,
+               url: cookbook_version_url(release.cookbook, release.version),
+               published: release.created_at do |entry|
+      entry.title "#{release.name} #{release.version}"
+      entry.content <<~CONTENT, type: 'html'
+        #{release.owner.name}: #{release.name} #{release.version} released by #{release.published_by.name}
+        #{cookbook_atom_content(release)}
+      CONTENT
+
+      entry.author do |author|
+        author.name "#{release.published_by.name} (#{release.published_by.username})"
+        author.uri user_url(release.published_by)
+      end
+    end
+  end
+end
--- a/src/supermarket/app/views/cookbook_versions/index.atom.builder
+++ b/src/supermarket/app/views/cookbook_versions/index.atom.builder
@@ -1,6 +1,7 @@
 atom_feed language: 'en-US' do |feed|
   feed.title 'Cookbook Releases'
   feed.updated safe_updated_at(@cookbook_versions)
+  feed.link rel_next_prev_link_tags @cookbook_versions
 
   @cookbook_versions.each do |release|
     feed.entry release,
--- a/src/supermarket/.rubocop.yml
+++ b/src/supermarket/.rubocop.yml
@@ -39,6 +39,7 @@ Metrics/BlockLength:
   Exclude:
     - Guardfile
     - Vagrantfile
+    - config/routes.rb
     - lib/tasks/**/*
     - spec/**/*
 Metrics/PerceivedComplexity:
--- a/src/supermarket/app/controllers/cookbook_versions_controller.rb
+++ b/src/supermarket/app/controllers/cookbook_versions_controller.rb
@@ -1,5 +1,16 @@
 class CookbookVersionsController < ApplicationController
-  before_action :set_cookbook_and_version
+  before_action :set_cookbook_and_version, except: :index
+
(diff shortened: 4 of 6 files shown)


src/supermarket/app/views/cookbook_versions/index.atom.builder
```diff
@@ -0,0 +1,22 @@
...
+  feed.updated safe_updated_at(@cookbook_versions)
+  feed.link rel_next_prev_link_tags @cookbook_versions
+
+  @cookbook_versions.each do |release|
+    feed.entry release,
+               url: cookbook_version_url(release.cookbook, release.version),
+               published: release.created_at do |entry|
+      entry.title "#{release.name} #{release.version}"
+      entry.content <<~CONTENT, type: 'html'
+        #{release.owner.name}: #{release.name} #{release.version} released by #{release.published_by.name}
```
pschaumburg: Hey Robb,

I'd change this line to something like: 
`#{release.published_by.name} released version #{release.version} of #{release.name}`

This only shows the publisher of a new cookbook version, removes the author (as this is pretty annoying when you're the only one, who uploads new cookbook versions) and changed the wording a bit.

pschaumburg:
Hi Robb,
in addition to the made comment, I'm missing some content of the Changelog, which should contain something like the current version number and the containing release notes (it should only show the changes for the specific matching cookbook version). 
Is this also possible? :) 

Best regards,
Patrick

src/supermarket/app/views/cookbook_versions/index.atom.builder
robbkidd: We could make this a little more complicated and show a different message for when the release is the owner versus when the releaser is a collaborator. Like:

### Owner is Releasing

> Sous Chefs: awesome_sauce 1.2.3 released

### Collaborator is Releasing

> Sauce Chefs: awesome_sauce 1.2.3 released by Hot Sauce Fan

What do you think?

robbkidd:
Regarding the changelog display, that could get tricky and to get this feed shipped, I'd recommend opening a different issue to talk about the expectations and desires for the changelogs.

src/supermarket/app/views/cookbook_versions/index.atom.builder
pschaumburg: That sounds like a great idea!
I'd really prefer that. Otherwise no one knows, who released it.

pschaumburg:
I agree with that. I'll create an issue for that when it has been delivered.

RajeshPaul38:
Need DCO. Looks good code wise.

src/supermarket/app/models/cookbook_version.rb
```diff
@@ -11,7 +11,8 @@ class CookbookVersion < ApplicationRecord
   has_many :metric_results, -> { includes :quality_metric }, inverse_of: :cookbook_version
 
   belongs_to :cookbook
-  belongs_to :user
+  belongs_to :user, -> { includes :chef_account }, inverse_of: :cookbook_versions
```
RajeshPaul38, line 14: Have we defined the `inverse_of` in user model as well. Because this will be useful only when it's declared in a bidirectional manner i.e. both `has_many`/`has_one` and `belongs_to`

RajeshPaul38:
@robbkidd  pls add DCO

RajeshPaul38:
This has merge conflicts. Pls resolve @robbkidd

robbkidd:
I have little inclination to update this PR.

All commits were made when I was an employee of Chef Software. Note the author/committer email on the commits is a `chef.io` address. Therefore, the code is Progress Chef's and the DCO is implied. If these changes are of interest to you, you have my permission (though, as mentioned, you do not need it) to include them as-is or modified any way you deem appropriate.

PR #1748, opened 2018-07-26
bring omnibus-internal cookbook up to Chef 14 standards

tas50:
There's no need for the sysctl cookbook. It's been folded into the chef-client and the current usage results in deprecation warnings.


tas50:
I'll need someone to actually test this out. It *should* work just fine, but I'm not sure how to Supermarket together to fully test it

omnibus/.kitchen.yml
```diff
@@ -9,7 +9,8 @@ driver:
 
 provisioner:
   name: chef_zero
-  require_chef_omnibus: 13.6.4
+  product_name: chef
+  product_version: 13.6.4
```
tas50: We should probably update this to 14 since that's what's being used.

robbkidd: Hmm. I am currently confused as to how Chef 14 is getting into the omnibus package. 

```
~$ sudo supermarket-ctl reconfigure
Starting Chef Client, version 14.3.37
resolving cookbooks for run list: ["omnibus-supermarket::default"]
...
```

And yet ...

```
# /opt/supermarket/version-manifest.txt
supermarket 3.1.74+20180815175347

Component                Installed Version       Version GUID                                                              Overridden From
---------------------------------------------------------------------------------------------------------------------------------------------
...
cacerts                  2018-01-17              sha256:defe310a0184a12e4b1b3d147f1d77395dd7a09e3428373d019bef5d542ceba3
chef-gem                 13.6.4                  11.12.2
config_guess             master                  git:84f04b02a7e2fc8eaa9d52deee5f6d57b06fe447
...
```

tas50: Isn't the kitchen file just for testing?

robbkidd: It is. This one is a kitchen for spinning up local omnibus builders and then performing the omnibus build. The Chef version in this one probably doesn't matter and could move to 14.

robbkidd: So ..uh.. yea, we can move it up. The version of Chef inside the omnibus package is a separate issue.

omnibus/cookbooks/omnibus-supermarket/metadata.rb
```diff
@@ -6,8 +6,9 @@
 long_description 'Installs/Configures supermarket in an Omnibus installation'
 version          '1.1.1'
 
+chef_version     '>= 14.0'
```
robbkidd: This it the bit that could cause some confusion. The omnibus installer is supposedly bringing in Chef 13 to support the `-ctl reconfigure` action.

robbkidd:
I rebased this one to bring it up to date with the flurry of recent changes.

robbkidd:
Added a couple fixup commits. Running another omnibus build and test with these. If successful, I'll rebase'n'squash.

tas50:
@robbkidd You're missing DCO on those last 2 commits

robbkidd:
Yup. Those commits will disappear when I autosquash during a rebase. Planned for the morning.

robbkidd:
```

robbkidd:
One more omnibus build, with _feeling_.

PR #1750, opened 2018-08-09
Set session cookie to secure when sent over an ssl connection

Vasu1105:
Currently When supermarket is being used over https, the secure flag should be set on the session cookie.
Fixes https://getchef.zendesk.com/agent/tickets/18535

Signed-off-by: Vasu1105 <vasundhara.jagdale@msystechnologies.com>


src/supermarket/config/initializers/session_store.rb
```diff
@@ -1,3 +1,3 @@
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.session_store :cookie_store, key: '_supermarket_session'
+Rails.application.config.session_store :cookie_store, key: '_supermarket_session', secure: Rails.env.production?
```
robbkidd: Sadly, not everyone running a "production" Supermarket instance is using HTTPS. 😢

This check should probably be something like `ENV['PROTOCOL'] == 'https'`. That's the environment variable that is set for the Rails process to know what URL scheme to use when creating links back to itself. The value can be either `"https"` ([default](https://github.com/chef/supermarket/blob/5911828fd3b5be367028c147e7962b520782b001/omnibus/cookbooks/omnibus-supermarket/attributes/default.rb#L333)) or `"http"` for Supermarket omnibus installs, all of which operate with `Rails.env.production? == true`.

robbkidd:
I tinkered a little bit to see how we could test this a little easier.

Because the configuration is being set in a Rails initializer, it will be difficult to test both cases for secure and non-secure session cookies through an expectation on `Supermarket::Application.config.session_options[:secure]`. So I propose that we put the logic of deciding whether to secure a session cookie in a method and call that. This is based around determining whether requests are coming in via HTTPS by looking at the value of `ENV['PROTOCOL']`.

# The Changes


## Add a new method to the `Supermarket::Host` class

`app/lib/supermarket/host.rb`:
```
    def self.secure_session_cookie?
      ENV['PROTOCOL'] == 'https'
    end
```

## Add tests

`spec/lib/supermarket/host_spec.rb`:
```
  describe "#secure_session_cookie?" do
    it "returns true if using SSL" do
      allow(ENV).to receive(:[]).with('PROTOCOL').and_return('https')
      expect(described_class.secure_session_cookie?).to be true
    end

    it "returns false if not using SSL" do
      allow(ENV).to receive(:[]).with('PROTOCOL').and_return('http')
      expect(described_class.secure_session_cookie?).to be false
    end
  end
```

Happy tests:

```
supermarket/src/supermarket > bundle exec rspec -f doc spec/lib/supermarket/host_spec.rb:34
[...]
Supermarket::Host
  #secure_session_cookie?
    returns false if not using SSL
    returns true if using SSL
  #full_url
    generates the correct url when there is no port present
    should not display the port in the url if it is 80 or 443
    generates the correct url when there is a port present

Finished in 0.48219 seconds (files took 5.99 seconds to load)
5 examples, 0 failures
```

## Update the session_store initializer to use this new method:

```
Rails.application.config.session_store :cookie_store,
  key: '_supermarket_session',
  secure: Supermarket::Host.secure_session_cookie?
```

# Manual Testing

## No SSL (default in development environment)

```
supermarket/src/supermarket > grep PROTOCOL .env.development
PROTOCOL=http

supermarket/src/supermarket > bundle exec rails console
Loading development environment (Rails 5.0.3)
[1] pry(main)> Supermarket::Application.config.session_options[:secure]
=> false
```

## With SSL (default in production omnibus installs)

```
supermarket/src/supermarket > grep PROTOCOL .env.development
PROTOCOL=https

supermarket/src/supermarket > bundle exec rails console
Loading development environment (Rails 5.0.3)
[1] pry(main)> Supermarket::Application.config.session_options[:secure]
=> true
```

Issue #1754, opened 2018-08-26
Celluloid is deprecated, git submodule missing, bundle install broken


rmoriz:
```
➜  supermarket git:(master) bundle
Fetching gem metadata from https://rubygems.org/.........
Fetching https://github.com/rubysec/bundler-audit.git
Fetching https://github.com/celluloid/celluloid
Submodule 'culture' (http://github.com/celluloid/culture.git) registered for path 'culture'
Cloning into '/Users/rmoriz/.asdf/installs/ruby/2.5.1/lib/ruby/gems/2.5.0/bundler/gems/celluloid-0d192022ffff/culture'...
remote: Repository not found.
fatal: repository 'https://github.com/celluloid/culture.git/' not found
fatal: clone of 'http://github.com/celluloid/culture.git' into submodule path '/Users/rmoriz/.asdf/installs/ruby/2.5.1/lib/ruby/gems/2.5.0/bundler/gems/celluloid-0d192022ffff/culture' failed
Failed to clone 'culture'. Retry scheduled
```

Celluloid is unmaintained and its former maintainer decided to delete http://github.com/celluloid/culture from github. 

supermarket uses a hard coded ref of the `celluloid` repo that still contains a submodule reference to the now deleted submodule:

https://github.com/chef/supermarket/blob/master/src/supermarket/Gemfile#L17-L18

Looks like someone (Chef Software?) either needs to sponsor the Celluloid maintainer or remove the dependency: https://github.com/celluloid/celluloid/issues/779

robbkidd:
A quick copy of some of what was said on #1755:

> Celluloid is a dependency of Sidetiq which is what Supermarket currently uses to handle scheduled background jobs in Sidekiq. Sidetiq's use of the Celluloid DSL breaks with this update.

> Sidetiq is also not maintained anymore. I believe the solution will be reassessing Supermarket's need for scheduled background jobs and, if determined that they are still needed, finding a replacement for scheduled jobs.

I am investigating alternatives.

PR #1753, opened 2018-08-25
fix schema metadata create/update

rmoriz:
`dateModified` and `datePublished` are the wrong way around here, causing wrong dates on google search results (since ~2014 I guess).

 For example:

ohai cookbook - chef supermarket

ohai cookbook - google-suche-1

PR #1755, opened 2018-08-26
update gems, update celluloid to latest ref

rmoriz:
- workaround for celluloid #1754 
- ignore CVE-2018-1000544 (rubyzip directory traversal) because has [no upstream fix yet ](https://github.com/rubyzip/rubyzip/issues/369) (it's required by `license_finder` gem)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ matrix:
         - cd $TRAVIS_BUILD_DIR/src/$NAME
         - bundle exec rake spec
         - bundle exec rubocop
-        - bundle exec bundle-audit check --update
+        - bundle exec bundle-audit check --update --ignore CVE-2018-1000544
 
     # Fieri Spec
     - env:
--- a/src/supermarket/Gemfile
+++ b/src/supermarket/Gemfile
@@ -14,8 +14,7 @@ gem 'omniauth-oauth2', '~> 1.3.1'
 gem 'sidekiq', '~> 4.2'
 # Required until to build from master until Celluloid 0.17.4 or 0.18.0 is released
 # https://github.com/celluloid/celluloid/issues/728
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -59,7 +59,7 @@ matrix:
         - cd $TRAVIS_BUILD_DIR/src/$NAME
         - bundle exec rake spec
         - bundle exec rubocop
-        - bundle exec bundle-audit check --update
+        - bundle exec bundle-audit check --update --ignore CVE-2018-1000544
```
tas50: @robbkidd I wonder if we can just take out bundle audit since github is doing the same thing for us now

robbkidd: 🤔 There's a part of me that likes having a vulnerability scan as a test that passes or fails, but I admit that could be just because that's the way I've been doing it for so long.

robbkidd:
@rmoriz I :heart: you for submitting this and catalyzing a fix for something we've been kicking down the road for waaaay too long. I'm going to close this PR and continue the discussion in the issue #1754.

rmoriz:
Thanks. Bare in mind that the PR was meant to get get https://github.com/chef/supermarket/pull/1753 green which should work besides the 2 issues (celluloid, bundler-audit).

rmoriz:
Oh, a quick fix would be to look for local copies of that http://github.com/celluloid/culture.git repo and provide it to the maintainer https://github.com/celluloid/celluloid/issues/779#issuecomment-416064730

robbkidd:
>  ... a quick fix would be to look for local copies of that http://github.com/celluloid/culture.git repo ...

Uuuh, about that ... I kinda wiped out my local gem cache to replicate the problem. I hadn't quite thought out the implications of "things don't work because the repo isn't on GitHub anymore."

tenor-246291829

In my defense, I hadn't had my coffee yet.

PR #1757, opened 2018-08-30
use a supported Ubuntu release

robbkidd:
The precise APT repo is no longer. Trusty is the release used by the workers at the moment, so let's use that.

Attempting to resolve:
```
Adding APT Sources (BETA)
Disallowing sources: chef-stable-precise
To add unlisted APT sources, follow instructions in https://docs.travis-ci.com/user/installing-dependencies#Installing-Packages-with-the-APT-Addon
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,7 +96,7 @@ matrix:
       addons:
         apt:
           sources:
-            - chef-stable-precise
+            - chef-stable-trusty
           packages:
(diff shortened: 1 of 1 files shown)


robbkidd:
Accepting the test failure on Supermarket to get at the test successes for foodcritiquing and cookstyling `supermarket-deploy` cookbook.

PR #1758, opened 2018-08-30
map changelog sections to PR labels

robbkidd:
Makes for more organized changelog entries.


robbkidd:
If @tduffield says it's good, :shipit:

PR #1759, opened 2018-08-30
ignore CVE-2018-1000544 (rubyzip)

robbkidd:
There is no fix released yet for the rubyzip vulnerability and it is only used by license_finder in the development environment.


robbkidd:
@rmoriz Change of plans, I've made a separate PR for the quieting down of the audit.

PR #1765, opened 2018-09-11
omit web server version from response headers

Nimesh-Msys:
Fixes MSYS-879

- By using 'server_tokens off' in the main nginx config file, version
  number of the server can be omitted by default for any server entries
  declared.
- Tests for version number omission were added to the internal cookbook

- [Similar changes were applied for chef-server](chef/chef-server#1514) by using
  'more_clear_headers', but that would require 'Nginx-More' module,
  hence we are using an nginx-only solution for the moment

### DIFF
--- a/omnibus/cookbooks/omnibus-supermarket/templates/default/nginx.conf.erb
+++ b/omnibus/cookbooks/omnibus-supermarket/templates/default/nginx.conf.erb
@@ -35,9 +35,8 @@ http {
   <% unless @nginx['disable_access_log'] -%>
   access_log	<%= @nginx['log_dir'] %>/access.log supermarket;
   <% end %>
-  <% if @nginx['server_tokens'] -%>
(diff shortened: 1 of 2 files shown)


Nimesh-Msys:
Got it @robbkid, my bad. I'll fix it

robbkidd:
@Nimesh-Msys I changed things just a little bit.

The single `server_tokens off` in `nginx.conf` is enough to turn off the version number by default for all `server` entries. So I removed the two entries from `rails.nginx.conf.erb`.

I added some Inspec tests for the `Server` header to be _only_ `"nginx"`. These test are run in the packaging pipeline after the omnibus build is compelte. If a version number sneaks back in somehow, these tests will catch it.

Nimesh-Msys:
Cool. Thanks @robbkidd  👍

PR #1766, opened 2018-09-11
update rubyzip

rhass:
Appease CVE check that finds [CVE-2017-5946](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-5946):

> The Zip::File component in the rubyzip gem before 1.2.1 for Ruby has a directory traversal vulnerability. If a site allows uploading of .zip files, an attacker can upload a malicious file that uses "../" pathname substrings to write arbitrary files to the filesystem.

rubyzip is brought into Supermarket as a dependency of license_finder, which is only present in a development environment. This CVE does not apply to the public Supermarket or to onprem installs using the Supermarket omnibus package.

### DIFF
--- a/omnibus/Gemfile.lock
+++ b/omnibus/Gemfile.lock
@@ -205,7 +205,7 @@ GEM
       rspec-core (>= 2, < 4, != 2.12.0)
     ruby-progressbar (1.10.0)
     rubyntlm (0.6.2)
-    rubyzip (1.2.1)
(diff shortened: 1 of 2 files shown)


robbkidd:
> This also updates all gem deps to the latest releases available.

That's a bold strategy, Cotton. 😉

I can dig into why the supermarket test is trying to load puma. We're unicorns around here.

PR #1768, opened 2018-10-23
add FIPS 140-2 support

robbkidd:
Three of the five services that comprise Supermarket are identified as appropriate for enabling FIPS mode on:

* front-facing reverse web proxy (nginx): in single-server standalone environments nginx is responsible for SSL termination
* web application service (rails): some crypto is performed by the web service for user authentication via hashing secrets
* background jobs (sidekiq): no crypto for secrets is performed in background jobs as of this writing, but the jobs share a code base and runtime configuration with the web service, so enabling it is no more difficult

# Changes

## Configuration options

+ new `['supermarket']['fips_enabled']` item added to the omnibus install to explicitly enable or disable FIPS mode; the default is set to nil which the omnibus install will treat as "determine FIPS mode based on whether FIPS is enabled in the kernel"
+ new `['supermarket']['ssl']['fips_ciphers']` added to provide overridable default ciphers when operating in FIPS mode

## Service Behavior

The service startup scripts for nginx, rails, and sidekiq are updated to `export OPENSSL_FIPS=1` per the OpenSSL documentation for enabling FIPS mode when `['supermarket']['fips_enabled']` is true--whether explicitly or when detected as enabled in the kernel.

An initializer has been added to the web application such that the rails and sidekiq services react to this environment variable by enabling `fips_mode` in the loaded Ruby OpenSSL library.

## Test

The `install-check` Inspec tests are updated with a FIPS section to test that an environment variable `OPENSSL_FIPS` has been set to `1` for the three services and that FIPS ciphers are set in the nginx SSL configuration.

### DIFF
--- a/omnibus/cookbooks/omnibus-supermarket/libraries/config.rb
+++ b/omnibus/cookbooks/omnibus-supermarket/libraries/config.rb
@@ -58,18 +58,68 @@ class Supermarket
       node.consume_attributes('supermarket' => secrets)
     end
 
-    def self.audit_config(node)
+    def self.audit_config(config)
+      audit_s3_config(config)
+      audit_fips_config(config)
+    end
+
+    def self.audit_s3_config(config)
       required_s3_vars = %w(s3_bucket s3_access_key_id s3_secret_access_key s3_region).freeze
-      any_s3_settings = required_s3_vars.any? { |key| !(node[key].nil? || node[key].empty?) }
-      all_s3_settings = required_s3_vars.all? { |key| !(node[key].nil? || node[key].empty?) }
+      any_s3_settings = required_s3_vars.any? { |key| !(config[key].nil? || config[key].empty?) }
+      all_s3_settings = required_s3_vars.all? { |key| !(config[key].nil? || config[key].empty?) }
 
       if any_s3_settings && !all_s3_settings
         raise IncompleteConfig, "Got some, but not all, of the required S3 configs. Must provide none or all of #{required_s3_vars} to configure cookbook storage in an S3 bucket."
       end
 
-      if node['s3_bucket'] =~ /\./ &&
-         (node['s3_domain_style'] != ':s3_path_url' || node['s3_region'] != 'us-east-1')
+      if config['s3_bucket'] =~ /\./ &&
+         (config['s3_domain_style'] != ':s3_path_url' || config['s3_region'] != 'us-east-1')
         raise IncompatibleConfig, "Incompatible S3 bucket settings. If the bucket name contains periods, the bucket must be in us-east-1 and the domain style must be :s3_path_url.\nAmazon recommends against periods in bucket names. See: https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html"
+      end
+    end
+
+    def self.audit_fips_config(config)
+      unless built_with_fips?(config['install_directory'])
+        if fips_enabled_in_kernel?
+          raise IncompatibleConfig, 'Detected FIPS is enabled in the kernel, but FIPS is not supported by this installer.'
+        end
+        if config['fips_enabled']
+          raise IncompatibleConfig, 'You have enabled FIPS in your configuration, but FIPS is not supported by this installer.'
+        end
+      end
+    end
+
+    def self.built_with_fips?(install_directory)
+      File.exist?("#{install_directory}/embedded/lib/fipscanister.o")
+    end
+
+    def self.fips_enabled_in_kernel?
+      fips_path = '/proc/sys/crypto/fips_enabled'
+      (File.exist?(fips_path) && File.read(fips_path).chomp != '0')
+    end
+
+    def self.maybe_turn_on_fips(node)
+      # the compexity of this method is currently needed to figure out what words to display
+      # to the poor human who has to deal with FIPS
+      case node['supermarket']['fips_enabled']
+      when nil
+        # the default value, set fips mode based on whether it is enabled in the kernel
+        node.normal['supermarket']['fips_enabled'] = Supermarket::Config.fips_enabled_in_kernel?
+        if node['supermarket']['fips_enabled']
+          Chef::Log.warn('Detected FIPS-enabled kernel; enabling FIPS 140-2 for Supermarket services.')
+        end
+      when false
+        node.normal['supermarket']['fips_enabled'] = Supermarket::Config.fips_enabled_in_kernel?
+        if node['supermarket']['fips_enabled']
+          Chef::Log.warn('Detected FIPS-enabled kernel; enabling FIPS 140-2 for Supermarket services.')
+          Chef::Log.warn('fips_enabled was set to false; ignoring this and setting to true or else Supermarket services will fail with crypto errors.')
+        end
+      when true
+        Chef::Log.warn('Overriding FIPS detection: FIPS 140-2 mode is ON.')
+      else
+        node.normal['supermarket']['fips_enabled'] = true
+        Chef::Log.warn('fips_enabled is set to something other than boolean true/false; assuming FIPS mode should be enabled.')
+        Chef::Log.warn('Overriding FIPS detection: FIPS 140-2 mode is ON.')
       end
     end
 
--- a/omnibus/cookbooks/omnibus-supermarket/recipes/config.rb
+++ b/omnibus/cookbooks/omnibus-supermarket/recipes/config.rb
@@ -35,6 +35,7 @@ Supermarket::Config.load_or_create_secrets!(
 )
 
 Supermarket::Config.audit_config(node['supermarket'])
+Supermarket::Config.maybe_turn_on_fips(node)
 
 # Copy things we need from the supermarket namespace to the top level. This is
 # necessary for some community cookbooks.
--- a/omnibus/cookbooks/omnibus-supermarket/recipes/rails.rb
+++ b/omnibus/cookbooks/omnibus-supermarket/recipes/rails.rb
@@ -49,6 +49,7 @@ template 'rails.nginx.conf' do
   variables(nginx: node['supermarket']['nginx'],
             rails: node['supermarket']['rails'],
             fqdn: node['supermarket']['fqdn'],
+            fips_enabled: node['supermarket']['fips_enabled'],
             ssl: node['supermarket']['ssl'],
             app_directory: node['supermarket']['app_directory'])
 end
--- a/omnibus/cookbooks/omnibus-supermarket/spec/libraries/config_spec.rb
+++ b/omnibus/cookbooks/omnibus-supermarket/spec/libraries/config_spec.rb
@@ -24,6 +24,23 @@ describe Supermarket::Config do
   end
 
   describe '#audit_config' do
+    before(:each) do
+      allow(described_class).to receive(:audit_s3_config)
+      allow(described_class).to receive(:audit_fips_config)
+    end
+
+    it 'checks that the S3 configuration is valid' do
+      expect(described_class).to receive(:audit_s3_config).with(some: 'stuff')
+      described_class.audit_config(some: 'stuff')
+    end
+
+    it 'checks that the FIPS configuration is valid' do
+      expect(described_class).to receive(:audit_fips_config).with(some: 'stuff')
+      described_class.audit_config(some: 'stuff')
+    end
+  end
+
+  describe '#audit_s3_config' do
     let(:all_required_settings) do
       {
         's3_bucket' => 'bettergetabucket',
@@ -33,13 +50,13 @@ describe Supermarket::Config do
       }
     end
     it 'passes if all required S3 config values are present' do
-      expect { described_class.audit_config(all_required_settings) }
+      expect { described_class.audit_s3_config(all_required_settings) }
         .not_to raise_error
     end
 
     it 'fails the chef run if S3 config is incomplete' do
       incomplete_s3_config = { 's3_bucket' => 'bettergetabucket' }
-      expect { described_class.audit_config(incomplete_s3_config) }
+      expect { described_class.audit_s3_config(incomplete_s3_config) }
         .to raise_error(Supermarket::Config::IncompleteConfig)
     end
 
@@ -57,7 +74,7 @@ describe Supermarket::Config do
             's3_region' => s3_region,
             's3_domain_style' => s3_domain_style
           )
-          expect { described_class.audit_config(ok_config) }
+          expect { described_class.audit_s3_config(ok_config) }
             .not_to raise_error
         end
       end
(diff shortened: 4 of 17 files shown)


robbkidd:
It might cut down on the change noise to review the last commit by itself: [The FIPSification of Supermarket](a5ebc7d).

omnibus/cookbooks/omnibus-supermarket/test/integration/default/inspec/controls/install-check.rb
```diff
@@ -31,7 +31,23 @@
...
+  %w(nginx rails sidekiq).each do |service_name|
+    describe file("/opt/supermarket/service/#{service_name}/run") do
+      its(:content) { should match(/export OPENSSL_FIPS=1/) }
+    end
+  end
+
+  describe file('/var/opt/supermarket/nginx/etc/sites-enabled/rails') do
+    its(:content) { should match(/ssl_ciphers FIPS/) }
+  end
 end
```
marcparadise: Can we add validations for the available ciphers as well?

robbkidd: Probably! Like, check the whole line `ssl_ciphers FIPS@STRENGTH:!aNULL:!eNULL` and not just cut it off at `FIPS`?

robbkidd: @marcparadise Re-reading this. Or do you mean adding a check for what the ciphers should be in the non-FIPS case?

omnibus/cookbooks/omnibus-supermarket/recipes/config.rb
```diff
@@ -40,6 +40,25 @@
 # necessary for some community cookbooks.
 node.consume_attributes('runit' => node['supermarket']['runit'])
 
+case node['supermarket']['fips_enabled']
```
robbkidd: @cattywampus This case statement to coerce fips_enabled to a boolean value is my latest attempt to address someone putting a string value instead. What do you think?

src/supermarket/config/initializers/fipsify.rb
```diff
@@ -0,0 +1,15 @@
...
+  OpenSSL.fips_mode = true
+
+  # Most of Ruby's standard Digest algorithms will error when used
+  # in FIPS mode because they are not FIPS-approved.
+  # Reroute calls to Digest's hash algorithms to use OpenSSL's algorithms
+  # which, when compiled for FIPS, are approved.
+  hashes_uses_by_rails = %w[SHA1 SHA256 SHA512]
+  hashes_uses_by_rails.each do |algorithm|
+    Digest.send(:remove_const, algorithm) if Digest.const_defined?(algorithm)
+    Digest.const_set(algorithm, OpenSSL::Digest(algorithm))
```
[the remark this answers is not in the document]
cattywampus: Haha, it's ok. I don't feel that bad about it either. The `digest` library will get you into trouble and swapping in the `OpenSSL::Digest` class has always help me get around it. 👍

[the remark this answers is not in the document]
robbkidd: I will mention that I know `Digest::MD5`—not swapped out for OpenSSL's MD5—is still used in some places in the Rails stack and _does not error when used_ on a FIPS-enabled host.

robbkidd:
Change the `Chef::Log` calls from `info` to `warn` so that the statements would actually appear in a standard run of `supermarket-ctl reconfigure`.

With INFO

```
[vagrant@default-centos-7 ~]$ sudo supermarket-ctl reconfigure
Starting Chef Client, version 14.6.47
OpenSSL FIPS 140 mode enabled
resolving cookbooks for run list: ["omnibus-supermarket::default"]
Synchronizing Cookbooks:
  - [...]
Installing Cookbook Gems:
Compiling Cookbooks...
Converging 75 resources
Recipe: omnibus-supermarket::config
[...]
```

With WARN

```
[vagrant@default-centos-7 ~]$ sudo supermarket-ctl reconfigure
Starting Chef Client, version 14.6.47
OpenSSL FIPS 140 mode enabled
resolving cookbooks for run list: ["omnibus-supermarket::default"]
Synchronizing Cookbooks:
  - [...]
Installing Cookbook Gems:
Compiling Cookbooks...
[2018-10-30T16:59:28+00:00] WARN: Detected FIPS-enabled kernel; enabling FIPS 140-2 for Supermarket services.
Converging 75 resources
Recipe: omnibus-supermarket::config
[...]
```

robbkidd:
As of this build, I've logged in and poked around most of the web forms on the test instance. I've run every `knife supermarket` command against it.

✅ `knife supermarket download COOKBOOK [VERSION] (options)`
✅ `knife supermarket install COOKBOOK [VERSION] (options)`
✅ `knife supermarket list (options)`
✅ `knife supermarket search QUERY (options)`
✅ `knife supermarket share COOKBOOK [CATEGORY] (options)`
✅ `knife supermarket show COOKBOOK [VERSION] (options)`
✅ `knife supermarket unshare COOKBOOK`

robbkidd:
Trying to test more.

I attempted, succeeded, but ultimately abandoned some specs to check that the `Digest::HASH`es were equal to the `OpenSSL::Digest::HASH`es. The specs worked, but would need to be run separately (i.e. with `--tag fips`) because the algorithm swap out occurs for the entire Ruby process for the tests. Could _not_ test that `OpenSSL.fips_mode == true` because none of our test environments have FIPS-enabled kernels or an OpenSSL compiled with the FIPS Object Module.

Still worth considering is what more might be able or appropriate to add to the new FIPS section of the Inspec profile that drives `supermarket-ctl test`.

robbkidd:
I did kind of like the code that came out of trying to write specs, though. Take a peek at 9d98a89 where the business logic of FIPSifying is wrapped up in a class and some methods with meaningful names.

robbkidd:
Added cookstyle, foodcritic, and chefspec to the Travis matrix. I'll address the plethora of cookstyle findings after the feature-required code changes have settled out through testing.

omnibus/cookbooks/omnibus-supermarket/libraries/config.rb
```diff
@@ -58,21 +58,71 @@ def self.load_or_create_secrets!(filename, node)
...
          raise IncompatibleConfig.new "Incompatible S3 bucket settings. If the bucket name contains periods, the bucket must be in us-east-1 and the domain style must be :s3_path_url.\nAmazon recommends against periods in bucket names. See: https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html"
       end
     end
 
+    def self.audit_fips_config(config)
+      if !built_with_fips?(config['install_directory'])
+        if fips_enabled_in_kernel?
+          raise IncompatibleConfig.new "Detected FIPS is enabled in the kernel, but FIPS is not supported by this installer."
+        end
+        if config['fips_enabled']
```
marcparadise: These checks imply that it's an error when fips is supported in kernel, but is not enabled in config?  Or can it run in non-fips mode on a FIPs kernel? (If so, the enable check + non-fips build may be incorrect.)

omnibus/cookbooks/omnibus-supermarket/libraries/config.rb
```diff
@@ -58,21 +58,71 @@ def self.load_or_create_secrets!(filename, node)
...
+      (File.exist?(fips_path) && File.read(fips_path).chomp != "0")
+    end
+
+    def self.maybe_turn_on_fips(node)
+      # the compexity of this method is currently needed to figure out what words to display
+      # to the poor human who has to deal with FIPS
+      case node['supermarket']['fips_enabled']
+      when nil
+        # the default value, set fips mode based on whether it is enabled in the kernel
+        node.normal['supermarket']['fips_enabled'] = Supermarket::Config.fips_enabled_in_kernel?
```
marcparadise: Looks like the audit/check to ensure it's enable in the build has already run at this point?

omnibus/cookbooks/omnibus-supermarket/libraries/config.rb
```diff
@@ -58,21 +58,71 @@ def self.load_or_create_secrets!(filename, node)
...
+      when false
+        node.normal['supermarket']['fips_enabled'] = Supermarket::Config.fips_enabled_in_kernel?
+        if node['supermarket']['fips_enabled']
+          Chef::Log.warn('Detected FIPS-enabled kernel; enabling FIPS 140-2 for Supermarket services.')
+          Chef::Log.warn('fips_enabled was set to false; ignoring this and setting to true or else Supermarket services will fail with crypto errors.')
+        end
+      when true
+        Chef::Log.warn('Overriding FIPS detection: FIPS 140-2 mode is ON.')
+      else
+        node.normal['supermarket']['fips_enabled'] = true
```
marcparadise: Can we check for tihs condition in audit instead?

omnibus/cookbooks/omnibus-supermarket/recipes/cookbook_test.rb
```diff
@@ -26,13 +26,13 @@
...
 version_to_install = node['supermarket']['test']['version_to_install'] || 'nope_set_VERSION_TO_INSTALL'
 
-package 'supermarket' do
-  case node['platform_family']
-  when 'debian'
-    provider Chef::Provider::Package::Dpkg
+case node['platform_family']
+when 'debian'
+  dpkg_package 'supermarket' do
     source "#{package_path}/supermarket_#{version_to_install}-1_amd64.deb"
```
marcparadise: Could we determine the package name and use the `package` resource instead of invoking the platform-specific resources?

omnibus/cookbooks/omnibus-supermarket/recipes/rails.rb
```diff
@@ -30,32 +30,36 @@
...
   owner node['supermarket']['user']
   group node['supermarket']['group']
   mode '0600'
   variables(nginx: node['supermarket']['nginx'],
             rails: node['supermarket']['rails'],
             fqdn: node['supermarket']['fqdn'],
+            fips_enabled: node['supermarket']['fips_enabled'],
             ssl: node['supermarket']['ssl'],
             app_directory: node['supermarket']['app_directory'])
-  notifies :reload, 'runit_service[nginx]' if node['supermarket']['nginx']['enable']
```
marcparadise: Why the general change to shift to `subscribes` for the service to these files, instead of `notifies` here?

omnibus/cookbooks/omnibus-supermarket/spec/libraries/config_spec.rb
```diff
@@ -75,10 +92,146 @@
...
+  describe '#audit_fips_config' do
+    let(:default_config) do
+      { 'fips_enabled' => nil }
+    end
+
+    context 'when the installer was built to support FIPS' do
+      it 'does not raise' do
+        expect(described_class).to receive(:built_with_fips?).and_return(true)
+
+        expect{ described_class.audit_fips_config(default_config) }
```
marcparadise: Nit: I prefer to avoid explicit "this didn't happen" expectations - could we instead invoke `audit_fips_config` and rely on a raised exception to cause the test to fail?

omnibus/cookbooks/omnibus-supermarket/spec/libraries/config_spec.rb
```diff
@@ -75,10 +92,146 @@
...
+        it 'enables FIPS anyway and logs messages about why you gotta FIPS when it is enabled in the kernel' do
+          expect(Chef::Log).to receive(:warn).with('Detected FIPS-enabled kernel; enabling FIPS 140-2 for Supermarket services.')
+          expect(Chef::Log).to receive(:warn).with('fips_enabled was set to false; ignoring this and setting to true or else Supermarket services will fail with crypto errors.')
+          described_class.maybe_turn_on_fips(node)
+          expect(node['supermarket']['fips_enabled']).to be true
+        end
+      end
+    end
+
+    context 'with fips_enabled set to anything other than nil or boolean true/false' do
```
marcparadise: I didn't comment in the check above, but I am wondering why were' supporting non-boolean values with a default behavior? It seems opaque to the user.

omnibus/cookbooks/omnibus-supermarket/spec/recipes/app_spec.rb
```diff
@@ -1,9 +1,8 @@
+require 'spec_helper'
+
 describe 'omnibus-supermarket::app' do
-  let(:chef_run) do
-    ChefSpec::SoloRunner.new(platform: 'ubuntu', version: '16.04') do |node|
-      node.automatic['memory']['total'] = '16000MB'
-    end.converge(described_recipe)
-  end
+  platform 'ubuntu', '16.04'
```
marcparadise: I'm not familiar with `platform`, `automatic_attributes` without the context of `chef_run`; how do these get picked up?

omnibus/cookbooks/omnibus-supermarket/test/integration/default/inspec/controls/install-check.rb
marcparadise: Checking for the nginx list of ciphers (including FIPs) is correct in the FIPs case; I think we're only checking the env var for nginx here.

omnibus/cookbooks/omnibus-supermarket/recipes/rails.rb
marcparadise: answer: 'cause it's a proper resource now

omnibus/cookbooks/omnibus-supermarket/libraries/config.rb
robbkidd: This is where the `audit` login was extracted. @cattywampus and I had a pretty gnarly chain of `if`/`else`s to figure out what to tell the administrator configuring Supermarket. So what is left is similar, but different.

`audit` is about stopping `supermarket-ctl reconfigure` if a non-viable config is detected and presenting the admin with an error at a time they can do something about it.

`maybe_turn_on_fips` is about coercing `node['supermarket']['fips_enabled']` to a proper `true`/`false` boolean for resources downstream in the resource collection run by this cookbook.  The gnarliness here remains mostly as a result of the words we would like to display in the `supermarket-ctl reconfigure` output explaining what was detected/set and why the final value of the node attribute may have changed from what the admin expected.

omnibus/cookbooks/omnibus-supermarket/recipes/rails.rb
robbkidd: In getting chefspec to pass again for this cookbook, it seems chefspec doesn't handle notifies targeting resources that are declared later. Also, these notifies were targeting `runit_service` resources that are declared inside the `component_runit_service` resource, which seems dirty. That's among the reasons why I [added these actions](https://github.com/chef-cookbooks/enterprise-chef-common/pull/67/files#diff-ee7d560240df9baed002da7d844d9576R108) to the `component_runit_service` custom resource in chef-cookbooks/enterprise-chef-common#67.

omnibus/cookbooks/omnibus-supermarket/spec/libraries/config_spec.rb
robbkidd: This came out of some exploring admin expectations and some prior art with FIPSing in core Chef.

@cattywampus and I were confused at first when testing this in a Test Kitchen VM. We had set fips_enabled to `true` through TK's attributes and still weren't seeing the Supermarket services turn FIPS on. Turns out it was the explicit check for boolean `node['supermarket']['fips_enabled'] == true` in the service templates and TK was setting it to the String `"true"`.

Looking at how FIPS is detected, disabled, or enabled shows that we have previously used "literally anything else" as a way to enable FIPS, so we did that. Downstream code in the cookbook can rely on `fips_enabled` being a Proper Boolean while we can accommodate admins giving a string form of truth to the config.

omnibus/cookbooks/omnibus-supermarket/spec/recipes/app_spec.rb
robbkidd: This is the awesomeness of ChefSpec 7.

omnibus/cookbooks/omnibus-supermarket/recipes/cookbook_test.rb
marcparadise: Aside: seems like that could be a good change to the package resource - allowing it to auto resolve extension if source is provided without one.; Or to accept a platform hash for `source`.   

In any case, no need to change it - t's mostly a minor frustration on my part. We have so much package handling across our products that has to work around this, it seems something that the abstraction should provide.

omnibus/cookbooks/omnibus-supermarket/spec/libraries/config_spec.rb
marcparadise: Right, I mean having the test consist of:
`described_class.audit_fips_config` without the surrounding `expect`. If it does erroneously raise, the test will fail without using `not_raise`

omnibus/cookbooks/omnibus-supermarket/spec/libraries/config_spec.rb
marcparadise: Ugh. Okay thanks.

omnibus/cookbooks/omnibus-supermarket/recipes/cookbook_test.rb
robbkidd: Yea. Another reason to not fiddle with this too much is this recipe is only used in the Test Kitchen testing environment, not in any `supermarket-ctl reconfigure` in production.

I'm willing to poke at this to lessen the frustration, though.

omnibus/cookbooks/omnibus-supermarket/spec/libraries/config_spec.rb
robbkidd: Oooooh. 🤔 Do you think the "it does not raise" statement is sufficient for establishing the expectation for the reader? I tend to skim a spec looking for `expect` statements to find out what's being tested.

omnibus/cookbooks/omnibus-supermarket/libraries/config.rb
robbkidd: This check-and-assignment is done here for the same reason as [explained on a different comment](https://github.com/chef/supermarket/pull/1768#discussion_r): this method `maybe_turn_on_fips` is about coercing `node['supermarket']['fips_enabled']` to a proper true/false boolean while `audit` is about raising errors for non-viable configurations.

omnibus/cookbooks/omnibus-supermarket/spec/libraries/config_spec.rb
robbkidd: I'll simplify this down to just the method call.

omnibus/cookbooks/omnibus-supermarket/recipes/cookbook_test.rb
robbkidd: I have rediscovered why `dpkg_package` and `rpm_package` were used instead of `package`: Ubuntu (and probably Debian) platforms default to the APT package provider and APT doesn't take a source filename to install.

```
           Chef::Exceptions::Package
           -------------------------
           apt package provider cannot handle source attribute. Use dpkg provider instead

           Resource Declaration:
           ---------------------
           # In /tmp/kitchen/cache/cookbooks/omnibus-supermarket/recipes/cookbook_test.rb

            35: package 'supermarket' do
            36:   source package_path_and_filename
            37: end
            38:
```

robbkidd:
Rebased against the branch for #1788. Review and merge that one first.

sclark007:
Thanks @cattywampus and @robbkidd this is awesome!!

PR #1770, opened 2018-10-23
reduce package size with compression for omnibus DEB & RPM configs

tas50:
For other projects this reduced package sizes by about 23%


robbkidd:
oooo

robbkidd:
Merging this without the Travis status because I've apparently confused Travis with some force pushes and this change does not effect anything our Travis config tests for anyway.

PR #1773, opened 2018-10-23
Update Foodcritic to 14.3

tas50:


tas50:
I was wondering where fieri ended up

PR #1779, opened 2018-11-01
 fix itemprop schema for cookbook create/update datetimes

robbkidd:
(Opening a new PR for @rmoriz because #1753 has gotten very confused.)

[`dateModified`](https://schema.org/dateModified) and [`datePublished`](https://schema.org/dateCreated) are the wrong way around in the cookbook view HTML, causing wrong dates on google search results (since ~2014).

For example:

ohai cookbook - chef supermarket

ohai cookbook - google-suche-1

chef cookbook openssl - google-suche

### DIFF
--- a/src/supermarket/app/views/cookbooks/_sidebar.html.erb
+++ b/src/supermarket/app/views/cookbooks/_sidebar.html.erb
@@ -52,8 +52,8 @@
     </div>
 
     <h4>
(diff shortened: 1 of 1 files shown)


robbkidd:
@chef/supermarket-maintainers, assemble! I require one of your thumbs on this!

PR #1780, opened 2018-11-01
remove unnecessary build cruft before packaging (reduces size ~12%)

tas50:
In other projects we've seen up to a 30% reduction in the install size with this change.


robbkidd:
```
268M Dec 12 22:35 supermarket_3.1.96-1_amd64.deb
206M Dec 20 18:06 supermarket_3.2.0-1_amd64.deb
```

PR #1784, opened 2018-11-28
update Rails to 5.0.7.1

robbkidd:
Updates activejob to 5.0.7.1 to address [CVE-2018-16476](https://groups.google.com/forum/#!topic/rubyonrails-security/FL4dSdzr2zw). Supermarket does not use activejob for background workers and so is not vulnerable to this CVE. Updating to keep-up-to-date and quiet CVE audits.


robbkidd:
INORITE?

PR #1785, opened 2018-11-29
update rack (& other gems) in the omnibus build environment

robbkidd:
Quiets the CVE audit in GitHub. Rack isn't used in the omnibus build and so presents no vulnerabilities related to CVE-2018-16470.

Also attempts an update of all the gems in the build environment.


robbkidd:
Holding off on merging this until the pipeline is green on an ad-hoc build.

PR #1786, opened 2018-11-30
Add support for OCSP in Nginx

safematix:
Implementation of support for OCSP in Nginx.


robbkidd:
@safematix Thanks for this! Can you share more about what problem this solve for you? What would you recommend be added to the configuration documentation about why an administrator of a public or private Supermarket would want to turn this on?

robbkidd:
@safematix Can you share more about the problem this solves for you? And any recommendations for documenting when and how other Supermarket administrators would want to enable this feature in their environment? I don't know enough about OCSP to to carry this change to the finish line. Without more information, this PR will be closed in a week or so.

robbkidd:
Closing because of inactivity. Can be reopened for consideration if someone is interested in fleshing out the requested details of this change.

PR #1788, opened 2018-12-04
update omnibus to use latest enterprise cookbook

robbkidd:
Use the enterprise cookbook from public Supermarket. Remove it as a dependency from cookbooks that aren't actually using it!

---

My current thinking is to keep the Appeasement commits for future people to know why a line changed in the git history. I think I'd like to avoid muddying the archaeology by associating those style changes directly with the enterprise cookbook updating.

Because this is a suuuuuper noisy set of changes, the meat lives in f09611291dc3e3946a216637abb95888b3899f5f and shouldn't look much different than the changes reviewed over on #1768.

---

With the enterprise cookbook getting updated for component_runit_service to be a custom resource, the notifications to 'runit_service[x]' would error in ChefSpec and -ctl reconfigure runs. The enterprise cookbook has also updated to support passing a subset of actions on component_runit_service to the inner runit_service. However, notifies declared on resources before a target component_runit_service appears were failing in ChefSpec runs.

Therefore, here the resource notifications are flipped to wiring up subscribes on the component_runit_service resources that have within them the runit_service intended for restart/reload/etc.

For example, the nginx component_runit_service now subscribes to the 2 template resources responsible for NGINX configuration instead of those templates notifying a runit_service.

Some of the resources that previously notified a runit service have been updated to have simpler name properties so that the subscribes entries do not need the details from node attributes to identify resources to which to subscribe.

ALSO modernize the cookbook's specs to use ChefSpec's fancy new DSL for platform and node attribute values. Makes things slightly more readable.


robbkidd:
My current thinking is to keep the Appeasement commits for future people to know why a line changed in the git history. I think I'd like to avoid muddying the archaeology by associating those style changes directly with the enterprise cookbook updating.

omnibus/cookbooks/omnibus-supermarket/recipes/cookbook_test.rb
```diff
@@ -26,13 +26,13 @@
...
-    provider Chef::Provider::Package::Dpkg
+case node['platform_family']
+when 'debian'
+  dpkg_package 'supermarket' do
     source "#{package_path}/supermarket_#{version_to_install}-1_amd64.deb"
-  when 'rhel'
-    provider Chef::Provider::Package::Rpm
+  end
+when 'rhel'
+  rpm_package 'supermarket' do
```
robbkidd: Foodcritic didn't like `package` with a `provider` property and said "Use the specific package resource!"

Gotta use the specific `{dpkg|rpm}_package` resources because the `package` resource will default to `{apt|yum}_package` which (at least `apt_package`) don't accept a `source` property. See https://github.com/chef/supermarket/pull/1768#discussion_r238826986 for a previous review of this change.

robbkidd:
:rocket::cake:

PR #1792, opened 2018-12-29
Remove Chef Provisioning from the tools page

tas50:
We're in the process of winding down Chef provisioning and the only
tools here were from Chef and stuart. I've deleted all of the entries
from Chef.

### DIFF
--- a/src/supermarket/app/models/tool.rb
+++ b/src/supermarket/app/models/tool.rb
@@ -1,7 +1,7 @@
 class Tool < ApplicationRecord
   include PgSearch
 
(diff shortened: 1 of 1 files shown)


tas50:
@robbkidd It's probably ok if we note it in the release notes

tas50:
Pinging again on this one. It's time to say goodbye to provisioning.

robbkidd:
If we remove it this way and someone has a private Supermarket with provisioning tool entries, at best those tools will disappear from view within their organization. At worst, their Supermarket instance starts throwing exceptions and returning 500s when trying to search or display any tools because of the presence of tool records whose type is now unknown.

PR #1793, opened 2019-01-07
Change Username Displayed in the appheader

marcotesch:
Changed Username which is displayed within the appheader.

As the User Model already has a .name property which is propagated if the first and last name are set else it will be .username as before.

### DIFF
--- a/src/supermarket/app/views/application/_appheader.html.erb
+++ b/src/supermarket/app/views/application/_appheader.html.erb
@@ -14,7 +14,7 @@
     </ul>
   <% else %>
     <%= gravatar_for current_user, size: 80 %>
-    <%= current_user.username %> <i class="fa fa-caret-down"></i>
+    <%= current_user.name %> <i class="fa fa-caret-down"></i>
     <ul class="userdropdown">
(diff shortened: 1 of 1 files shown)


robbkidd:
Hi, @marcotesch! Thank you for this contribution. Can you say more about why this change is useful for you? Displaying the current user's user name instead of their full name is pretty common in web site menus.

marcotesch:
Hi @robbkidd,

Well the behavior how the .name property is filled in the user model seems to be decent in two ways:

1. If the user is willing to provide a first or last name it is filled with the concatination of these two properties, which then would be more personal and friendly to use this property in the appheader which was the reason for me to make the change. (As this way is also quite common on websites and a "market" might be more personal?)

2. If the user is not willing to provide any personal information the .name property still falls back to the old behavior which seems to be sufficent in that case. 

P.S.:
3. If Users are federated in an organization through an AD in most cases they are always "greeted" by their IT Resources with there first and last name ;-), so for consistency reasons this would be nice to have it in the chef supermarket as well.

robbkidd:
@marcotesch Thanks for the explanation. After some conversations with other stakeholders, we've determined that we want to continue to display the username.

PR #1795, opened 2019-02-06
Create omnibus buildkite pipelines

jeremiahsnapp:
This creates buildkite pipelines that will be used for omnibus builds/tests. It does not actually replace the current build/test infrastructure. That will be done in a separate PR once we finish testing the new pipelines.

### DIFF
--- a/omnibus/omnibus-test.sh
+++ b/omnibus/omnibus-test.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+set -ueo pipefail
+
+channel="${CHANNEL:-unstable}"
+product="${PRODUCT:-supermarket}"
+version="${VERSION:-latest}"
+
+echo "--- Installing $channel $product $version"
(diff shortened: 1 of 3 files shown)


omnibus/omnibus-test.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+set -ueo pipefail
+
+channel="${CHANNEL:-unstable}"
+product="${PRODUCT:-chef-server}"
```
robbkidd: Because this script lives vendored into the Supermarket repo, should the `product` variable maybe default to `supermarket`? Or just _be_ `supermarket`?

Or keep this as close to a common script as possible so updates to it can be dropped in with minimal differences?

jeremiahsnapp: great catch! it should be changed to supermarket. i'll fix that

PR #1796, opened 2019-02-11
Replace Jenkins omnibus build/test pipeline with Buildkite

jeremiahsnapp:

### DIFF
--- a/.delivery/build_cookbook/recipes/provision.rb
+++ b/.delivery/build_cookbook/recipes/provision.rb
@@ -28,15 +28,6 @@ end
 # Let delivery-truck do the workflow release promotion
 include_recipe 'delivery-truck::provision'
 
-# Promote the omnibus artifact to current
-expeditor_artifactory workflow_change_project do
-  ref_type 'omnibus.version'
-  ref ::File.read(::File.join(workflow_workspace_repo, 'VERSION')).strip
-  channel 'current'
-  action :promote
-  only_if { workflow_stage?('union') }
-end
-
 # Stand up the Supermarket instance
 execute 'apply-terraform' do
   command 'make apply'
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -49,22 +49,19 @@ merge_actions:
         - "Expeditor: Skip Changelog"
         - "Expeditor: Skip All"
         - "Expeditor: ACC Only"
-  - built_in:trigger_omnibus_build:
+  - trigger_pipeline:omnibus/release:
       ignore_labels:
         - "Expeditor: Skip Omnibus"
         - "Expeditor: Skip All"
         - "Expeditor: ACC Only"
       only_if: built_in:bump_version
-  - built_in:update_acc:
-      only_if_labels:
-        - "Expeditor: ACC Only"
 
 # These actions are taken, in the order specified, when an Omnibus artifact is promoted
 # within Chef's internal artifact storage system.
 #
 # TODO: add promoted_to_unstable action to update changelog with modified omnibus components
 subscriptions:
-  - workload: artifact_published:unstable:supermarket:*
+  - workload: artifact_published:current:supermarket:*
     actions:
       - built_in:update_acc
   - workload: artifact_published:stable:supermarket:*
--- a/.delivery/build_cookbook/recipes/smoke.rb
+++ b/.delivery/build_cookbook/recipes/smoke.rb
@@ -18,15 +18,6 @@
 # limitations under the License.
 #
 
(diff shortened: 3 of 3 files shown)


.expeditor/config.yml
```diff
@@ -64,10 +64,8 @@ merge_actions:
 #
 # TODO: add promoted_to_unstable action to update changelog with modified omnibus components
 subscriptions:
-  - workload: artifact_published:unstable:supermarket:*
-    actions:
-      - built_in:update_acc
   - workload: artifact_published:stable:supermarket:*
     actions:
+      - built_in:update_acc
```
robbkidd: Supermarket's pipeline in ACC is (was?) about going from unstable to stable, so I don't think moving this action has the desired result.

jeremiahsnapp: Great catch @robbkidd. I fixed it so now `update_acc` is subscribed to promotion to `current`.

.delivery/build_cookbook/recipes/provision.rb
```diff
@@ -28,15 +28,6 @@
...
 include_recipe 'delivery-truck::provision'
 
-# Promote the omnibus artifact to current
-expeditor_artifactory workflow_change_project do
-  ref_type 'omnibus.version'
-  ref ::File.read(::File.join(workflow_workspace_repo, 'VERSION')).strip
-  channel 'current'
-  action :promote
-  only_if { workflow_stage?('union') }
-end
```
robbkidd: I _think_ without this, there is nothing in the pipeline to promote a build from unstable to current.

BuildKite handles what the ACC smoke test did in the Acceptance stage: the multiple platform builds and the even more multiple platform tests. But if the change in ACC is approved out of Acceptance, this is the resource that performs the promotion from `unstable` to `current` in Union.

schisamo: The Buildkite pipeline which executes the Omnibus build now handles the promotion to current. This promotion is automatically executed after some basic testing of the packages are done. That testing is represented by this script:
https://github.com/chef/supermarket/blob/master/omnibus/omnibus-test.sh

More details on at:
https://expeditor.chef.io/docs/integrations/omnibus/

robbkidd: A-ha! I thought _wrong_, then.

robbkidd: I'd like to talk about the purpose of Supermarket's ACC pipeline in a world where Expeditor is performing the promotion from unstable to current after a single test run. Right now for a build of Supermarket to land in current, it must

* pass that single local test run (Jenkins' build-test-promote to unstable)
* provision and deploy to ACC's Supermarket Acceptance environment
* pass the smoke tests in Acceptance (I believe those are different tests?)
* get Approved by a human out of Acceptance
* Union's Provision promotes to current

The ACC pipeline does not perform the promotion to stable for release. That is a command to Expeditor. With the smoke tests and human approval removed as a requirement for a build to land in current, what needs are met by what is left of the ACC pipeline activities?

tduffield: For the time being, to serve as an acceptance environment.

schisamo: Another option we could consider is make the promotion to `current` optional in the new Buildkite Omnibus build pipeline and continue to allow the `expeditor_artifactory` resource to promote to `current` in the `union/provision` phase of the workflow pipeline. This would allow things to continue to work int he same way as @robbkidd has come to expect.

robbkidd: It's not that I want to preserve old behavior because that's what I expect. I'm wondering how much old behavior can be torn down so it doesn't have to be maintained.

robbkidd:
Vigorous agreement achieved during video chat with @schisamo, @jeremiahsnapp, and myself.

What remains of the Supermarket ACC pipeline manages Supermarket deployments in the four standard environments (AURD). We'll follow up with folks internally to confirm whether or not those environments are in use or the pipeline itself is a dependency of others.

.delivery/build_cookbook/recipes/provision.rb
robbkidd: Vigorous agreement achieved during video chat with @schisamo, @jeremiahsnapp, and myself.

What remains of the Supermarket ACC pipeline manages Supermarket deployments in the four standard environments (AURD). We'll follow up with folks internally to confirm whether or not those environments are in use or the pipeline itself is a dependency of others.

robbkidd:
I plan to merge tomorrow morning.

PR #1797, opened 2019-03-01
Explicitly define expeditor promote config

jeremiahsnapp:

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -68,3 +68,11 @@ subscriptions:
     actions:
       - built_in:rollover_changelog
       - built_in:notify_chefio_slack_channels
+
+promote:
+  actions:
+    - built_in:promote_artifactory_artifact
+  channels:
+    - unstable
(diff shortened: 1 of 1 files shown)


robbkidd:
Does this cause new behavior? If so, what is new?

Does being explicit here solve a problem? If so, what's the problem?

jeremiahsnapp:
No new behavior. We're just explicitly defining these expeditor default values to be consistent across our server projects.

robbkidd:
Ah! So, this is an explicit ordering of the channels through which this project's artifacts get promoted instead of relying on [the default for projects that define a `product_key`](https://expeditor.chef.io/docs/reference/promote/#promotion-channels)?

jeremiahsnapp:
Yes, that's correct @robbkidd. These channels are already ordered this way by default. Now they are explicitly defined in the expeditor config.

robbkidd:
Bueno! Thanks for the explanation!

PR #1803, opened 2019-04-19
supermarket-ctl reconfigure will perform license-acceptance flow

tyler-ball:
## Description
We enforce the EULA check in omnibus projects via the `reconfigure` command

## Related Issue
N/A

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have updated the documentation accordingly.


omnibus/config/templates/supermarket-ctl/supermarket-ctl.erb
```diff
@@ -15,20 +16,32 @@
...
+    # Need to allow '--accept-license' to be passed to the reconfigure command
+    def initialize(name, service_commands=true)
+      super
+      @category_command_map["general"]["reconfigure"][:arity] = 2
+    end
+
+    # Override reconfigure to first perform license-acceptance flow
+    def reconfigure(*args)
+      require "license_acceptance/acceptor"
+      LicenseAcceptance::Acceptor.check_and_persist!("supermarket", "<%= version %>")
```
robbkidd: Because some of the `supermarket-ctl` commands use other licensed things—`reconfigure` uses chef-zero, `test` uses InSpec—should the acceptance for Supermarket include those things, too?

tyler-ball:
> I think it might be worth a discussion on what the requirements are to produce a non-EULA'd build of Supermarket before this and #1803 go in.
@robbkidd  Happy to have that discussion here. I think it would be similar to what is happening with Chef Infra Client right now - allow users to easily overwrite trademarks with non-trademarked terms.

robbkidd:
> [...] allow users to easily overwrite trademarks with non-trademarked terms.

Right. "Supermarket" isn't currently a trademark Chef has listed. So, maybe, alternative distro packaging of it wouldn't require renaming?

Are users accepting a license for Supermarket _and_ its embedded Infra Chef chef-client and InSpec? Or only for the embedded things?

Would an alternative distro package of Supermarket only need to be able to swap out the embedded chef-client and inspec to not have to accept the Chef EULA?

tyler-ball:
> Are users accepting a license for Supermarket and its embedded Infra Chef chef-client and InSpec? Or only for the embedded things?

They would be accepting the license for all 3. Each product is governed by a license (they are all the same license but must be accepted for each product) and we streamline this by allowing them to accept all 3 licenses at the same time.

I am not sure about your other answers. If a community distro has the same name how do we differentiate it from the original product? Not even in a legal sense but even for things like package managers.

robbkidd:
After some out-of-band conversations, Chef Software has made the decision continue to distribute the Supermarket package under Apache 2.0, not under the Chef EULA. As such, installing and using Supermarket will _not_ require a license acceptance because 1) in-and-of-itself Supermarket is only useful in the scope of using the rest of the ecosystem tools and 2) its use of chef-client and inspec was determined to be as ‘library’ usage to help setup and configuration testing.

Use of Supermarket will not require accepting a license for Supermarket, Chef Infra Client or InSpec.

omnibus/config/templates/supermarket-ctl/supermarket-ctl.erb
robbkidd: I definitely still like bringing the `supermarket-ctl` command up to date to Ruby Glory.

robbkidd: I think this changes to ...?

```suggestion
      LicenseAcceptance::Acceptor.check_and_persist!(provided: "accept-no-persist")
```

robbkidd:
We're already talking about the embedded chef-client run in a commit comment.

I suspect the other place we would add a non-persistent acceptance is in [the shellout to inspec](https://github.com/chef/supermarket/blob/27a9ae1d9099a138f2d252519450f7395618942e/omnibus/cookbooks/omnibus-supermarket/files/default/ctl-commands/test.rb#L15-L18) for the `supermarket-ctl test` command.

tyler-ball:
Closing this since we decided to leave EULA flow out of Supermarket

PR #1809, opened 2019-05-08
Update the omnibus build license to the Chef EULA

btm:
The resulting distribuions are licensed under the Chef EULA.

Signed-off-by: Bryan McLellan <btm@loftninjas.org>


omnibus/config/projects/supermarket.rb
```diff
@@ -18,8 +18,8 @@
 maintainer "Chef Supermarket Team <supermarket@chef.io>"
 homepage "https://supermarket.chef.io"
 
-license "Apache-2.0"
-license_file "../LICENSE"
+license "Chef EULA"
+license_file "CHEF-EULA.md"
```
tas50: don't we need the file here as well

btm: yeah, I failed at git.

robbkidd:
I think it might be worth a discussion on what the requirements are to produce a non-EULA'd build of Supermarket before this and #1803 go in.

robbkidd:
In looking over this and #1803, I think it'd be best to have this change come along with that change. Anyone opposed to me cherry-picking this commit and including it in #1803?

tyler-ball:
@robbkidd Github now allows you to change the base branch this PR is looking at. If you change that to `https://github.com/chef/supermarket/tree/license_acceptance` merging this will merge it into https://github.com/chef/supermarket/pull/1803

Issue #1845, opened 2020-03-02
Links opening in new tabs should use `rel="noopener"`


cattywampus:
# Describe the problem

A customer security scan of the on-premise Chef Supermarket revealed a finding where links that open up in new tabs expose the Supermarket content to be manipulated through the `window.opener` context. More information on this issue can be found here:

  * https://mathiasbynens.github.io/rel-noopener/
  * https://developers.google.com/web/tools/lighthouse/audits/noopener

## Software Version

Chef Supermarket 3.2.0
Red Hat Enterprise Linux 7.5

## Stacktrace

Not applicable

## Possible Solution

Add `rel="noopener"` to hyperlinks where the `target` is set to `_blank`. It may also be necessary to add `noreferrer` value to the `rel` attribute for resolving this security concern for users with older browsers.

cattywampus:
This was merged into master and shipped in release 3.3.20

PR #1846, opened 2020-03-02
Prevent unsafe links to cross-origin destinations

cattywampus:
## Description

When using `target="_blank"` to open a link to another page, it's possible for the JavaScript on this newly opened tab to access the the `window` of the initiating link, allowing content in the dom to be modified to something that could deceive or harm the user. The fix for this is to include the attribute `rel="noopener"` in the anchor tag alongside `target="_blank"`. This ensures that any new browsing context created by the hyperlink does not have the `window.opener` browsing context. More details about this issue can be found here:

https://mathiasbynens.github.io/rel-noopener/

According to the link above, support for the `noopener` value was added to these browser versions:

  * Chrome 49
  * Firefox 55
  * Desktop Safari 10.1
  * iOS Safari 10.3

It's possible this could be implemented in Microsoft Edge, but I was not able to confirm that. In order to support this fix on older versions of the browsers above you have to set the rel attribute to `noreferrer`. This disables the opener context but also disables the HTTP Referrer header as well which might have an effect on referrer tracking through legitimate links.

This was discovered through a customer security scan of an on-premise Chef Supermarket installation. The recommended solution is to support both `noopener` and `noreferrer`, however I think the decision to use the second value is subjective and should be decided by the product manager. So long as customer browsers are relatively up to date (possible big leap?) only supporting `noopener` for now should be sufficient.

Signed-off-by: Keith Walters <kwalters@taphere.com>

## Related Issue

Issue #1845 

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:

- [ ] I have run the pre-merge tests locally and they pass.


cattywampus:
Looks like the travis build failed because `bundle exec bundle-audit check --update` discovered a vulnerability with the current version of `nokogiri` and should be upgraded from `1.10.7` to `1.10.8`.

Should that upgrade be handled as part of this pull request or opened as a new pr?

cattywampus:
Looks like the travis build errors would be addressed by Issues #1840 and #1841.

robbkidd:
@cattywampus Hey! Thanks for the PR. The master branch has been updated with newer rake and nokogiri to address the CVE tests. ... And I see you've rebased while I was typing this!

Issue #1839, opened 2020-02-14
installs with multiple web hosts throw unexpected CSRF errors


robbkidd:
# The problem

As of Supermarket 3.3.12, [the fix for actually reading in `secrets.json` during `supermarket-ctl reconfigure`](https://github.com/chef/supermarket/pull/1832) will cause **Supermarket installations with multiple web front-ends to throw cross-site scripting forgery errors for authenticated web UI interactions**. Comparing the contents of `secrets.json` on the front-end hosts will show that each has a different `secret_key_base` value. This causes each web front-end to distrust cookies written by any of the other front-ends.

## Some Background

Rails applications use a secret named `secret_key_base` as "the input secret to the application's key generator, which in turn is used to create all MessageVerifiers/MessageEncryptors, including the ones that sign and encrypt cookies." 

The command `supermarket-ctl reconfigure` will create a `secrets.json` file if one is not already present. In Supermarket versions up to 3.3.14, reconfigure creates a `secret_key_base` key, sets the value to a randomly generated hexadecimal string, and then "consumes" `secrets.json` into the node object. Once the `secrets.json` file was created, subsequent runs of `reconfigure` _entirely ignored it_ and never read those keys into the `reconfigure` run again (a bug fixed in #1832). As a result, all current Supermarket installations are setting `secret_key_base` in either `supermarket.rb` or `supermarket.json`. Left behind on every Supermarket web host is a `secrets.json` file with a uniquely created `secret_key_base`.

Configuration starts with `supermarket.rb`, is overridden by what is found in `supermarket.json`, which is overridden again by what is found in `secrets.json`. Now that Supermarket is paying attention to `secrets.json` in every `reconfigure` run, the previously unique-per-host secret is winning which causes each web host to distrust anything hashed or signed by the other web hosts.

## Software Version

Versions 3.3.12 - 3.3.14
 
## Replication Case

A Supermarket install:
* running 3.3.12 to 3.3.14
* with two or more web application hosts

Attempts to login will fail. The URL in the browser location will mention `csrf`. 

## Possible Solution(s)

1. Manage the contents of `secrets.json` with the wrapper cookbook (or whatever deployment automation is in use) that is installing Supermarket services. Either touch the file and leave it empty or write to it the key/value pair for `secret_key_base` to use across all web front-end hosts.
[...]

PR #1849, opened 2020-03-02
generate secrets.json with warning messages

robbkidd:
## Description

As described in #1839, secrets.json now wins for config values.

If a secret_key_base has been provided by an existing secrets.json file,
no problem so long as that file is the same across multiple web hosts.

If no secrets.json file is present and a secret_key_base value was
provided via supermarket.rb or supermarket.json, a warning is
presented that this value should really be in secrets.json and
the secrets.json file is generated with that value and used from there.

If no secrets.json file is present and no secret_key_base value was
provided, a warning is presented that a value is being generated. If
the installation has only one web host, the generated value is fine. If
the installation has 2 or more web hosts, the administrators are
required to ensure that the secrets.json file matches across those
hosts.

The omnibus installer is not empowered to resolve a discrepancy in secrets
values across multiple hosts. That is a secrets management/orchestration
problem outside the purview of omnibus.

## Related Issue

#1839

## Types of changes

- ~New feature (non-breaking change which adds functionality)~
- ~Breaking change (fix or feature that would cause existing functionality to change)~
- ~Chore (non-breaking change that does not add functionality or fix an issue)~

## Checklist:

- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

### DIFF
--- a/omnibus/cookbooks/omnibus-supermarket/libraries/config.rb
+++ b/omnibus/cookbooks/omnibus-supermarket/libraries/config.rb
@@ -46,7 +46,15 @@ class Supermarket
       node.consume_attributes('supermarket' => secrets)
     rescue Errno::ENOENT
       begin
-        secrets = { 'secret_key_base' => SecureRandom.hex(50) }
+        secret_key_base = if node['supermarket'] && node['supermarket']['secret_key_base']
+                            Chef::Log.warn 'Using secret_key_base from supermarket.rb or .json. This value should really be managed in secrets.json. Writing to secrets.json.'
+                            node['supermarket']['secret_key_base']
+                          else
+                            Chef::Log.warn 'No secret_key_base set! Generating and writing one to secrets.json. If this Supermarket installation has multiple hosts, you must duplicate the secrets.json file exactly across all hosts.'
+                            SecureRandom.hex(50)
+                          end
+
+        secrets = { 'secret_key_base' => secret_key_base }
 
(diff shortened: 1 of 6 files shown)


robbkidd:
I think I'll spend a little more time seeing if some tests can be written around this. The config loading code has increased its gnarliness.

omnibus/cookbooks/omnibus-supermarket/libraries/config.rb
```diff
@@ -46,7 +46,15 @@ def self.load_or_create_secrets!(filename, node)
       node.consume_attributes('supermarket' => secrets)
     rescue Errno::ENOENT
       begin
-        secrets = { 'secret_key_base' => SecureRandom.hex(50) }
+        secret_key_base = if node['supermarket'] && node['supermarket']['secret_key_base']
```
teknofire: Could simplify this a bit by using `dig`
```
secrets_key_base = if node.dig('supermarket', 'secret_key_base')
```

robbkidd: Mmmm. `.dig`. Nice. 🤔

robbkidd: I tried `node.dig`. Tests passed in my development environment, but didn't in CI. I think I'm at the limit of my rabbit hole timebox for testing this, so Imma opt for ugly `&&` check.

PR #1857, opened 2020-04-30
Change log directory mode to use an attribute

### LINKED ISSUE #1856: We could allow users to set custom permissions on the log directories
The current recipe doesn't allow users to set group permissions.

# Describe the problem
If a user wants to allow access to the Supermarket log directory based on Group, they're currently unable to do so as `mode '0700' is [in the recipe](https://github.com/chef/supermarket/blob/f426d598a6cf62cb82dff3cbe20a31dbae9540e2/omnibus/cookbooks/omnibus-supermarket/recipes/config.rb#L67:L72).

## Software Version
all applicable versions

## Replication Case
no replication needed, 

## Stacktrace

```
Compiling Cookbooks...
Converging 60 resources
Recipe: omnibus-supermarket::config
  * linux_user[supermarket] action create (up to date)
  * group[supermarket] action create (up to date)
  * directory[/etc/supermarket] action create (up to date)
  * directory[/var/opt/supermarket] action create (up to date)
  * directory[/var/log/supermarket] action create
    - change mode from '0740' to '0700'
    - restore selinux security context
```

## Possible Solution
PR incoming

moutons:
* Users may want to set different permissions on the log directories
  than our default of 0700, using an attribute allows them to do so.

Signed-off-by: Shaun Mouton <smouton@chef.io>

## Description
If a user wants to allow a group to access the log directory, the current recipe doesn't allow them to do so. This updates the recipe to fix this, the default attribute is set to mode '0700', the same as the current line in each recipe, so no behavior change is introduced for folks who don't set this attribute.

## Related Issue
Fixes #1856 

## Types of changes
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:
- [ ] I have run the pre-merge tests locally and they pass.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

### DIFF
--- a/omnibus/cookbooks/omnibus-supermarket/attributes/default.rb
+++ b/omnibus/cookbooks/omnibus-supermarket/attributes/default.rb
@@ -72,6 +72,7 @@ default['supermarket']['config_directory'] = '/etc/supermarket'
 default['supermarket']['install_directory'] = '/opt/supermarket'
 default['supermarket']['app_directory'] = "#{node['supermarket']['install_directory']}/embedded/service/supermarket"
 default['supermarket']['log_directory'] = '/var/log/supermarket'
+default['supermarket']['log_directory']['mode'] = '0700'
 default['supermarket']['var_directory'] = '/var/opt/supermarket'
 default['supermarket']['data_directory'] = '/var/opt/supermarket/data'
 default['supermarket']['user'] = 'supermarket'
--- a/omnibus/cookbooks/omnibus-supermarket/recipes/config.rb
+++ b/omnibus/cookbooks/omnibus-supermarket/recipes/config.rb
@@ -67,7 +67,7 @@ end
 directory node['supermarket']['log_directory'] do
   owner node['supermarket']['user']
   group node['supermarket']['group']
-  mode '0700'
+  mode node['supermarket']['log_directory']['mode']
   recursive true
 end
 
--- a/omnibus/cookbooks/omnibus-supermarket/recipes/nginx.rb
+++ b/omnibus/cookbooks/omnibus-supermarket/recipes/nginx.rb
@@ -27,7 +27,7 @@ include_recipe 'omnibus-supermarket::config'
   directory dir do
     owner node['supermarket']['user']
     group node['supermarket']['group']
-    mode '0700'
+    mode node['supermarket']['log_directory']['mode']
     recursive true
   end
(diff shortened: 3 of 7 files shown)


omnibus/cookbooks/omnibus-supermarket/attributes/default.rb
```diff
@@ -72,6 +72,7 @@
 default['supermarket']['install_directory'] = '/opt/supermarket'
 default['supermarket']['app_directory'] = "#{node['supermarket']['install_directory']}/embedded/service/supermarket"
 default['supermarket']['log_directory'] = '/var/log/supermarket'
+default['supermarket']['log_directory']['mode'] = '0700'
```
robbkidd, line 75: Sadly, we won't be able to set a `['supermarket']['log_directory']['mode']` subkey when the parent key `['supermarket']['log_directory']` has a value type of `String`.

robbkidd, line 75: Sadly, we won't be able to set a ['supermarket']['log_directory']['mode'] subkey when the parent key ['supermarket']['log_directory'] has a value type of String.

moutons: sorry for the lack-of-sleep-induced unforced error
what about `default['supermarket']['log_directory_mode']`

robbkidd: Cooould do that.

Now that I've also woken up a bit more, I thought of another option:
1. No configurable mode, so no new attribute needed.
2. The hard-coded mode changes from 0700 to 0740.
3. User accounts that get access to read Supermarket logs should be joined to the supermarket group by something outside of the Supermarket omnibus installer's internal chef-client run.

robbkidd:
In a re-review, I think I like the option of a smaller change: update the hard-coded permission from from `'0700'` to `'0740'` and document that any user that needs read access to the logs should be added to the `supermarket` group.

Anyone opposed? I'm happy to do the work or help someone do the work to close out this issue!

antima-gupta:
@RajeshPaul38  @saghoshprogress can we close this PR?

PR #1878, opened 2020-08-06
Update the bundled Chef Infra Client to 16.3.45

tas50:
Avoids double gems and should drop the overall package size.


tas50:
This blows up the world

PR #1882, opened 2020-08-06
Bump Cookstyle / Chefstyle to the latest

tas50:
This updates to the latest RuboCop engine as well.


robbkidd:
I'm sorry. There's another Gemfile to sort out via fieri's gemspec down in https://github.com/chef/supermarket/blob/master/src/supermarket/engines/fieri/fieri.gemspec#L19 .

tas50:
I updated the second gemfile.lock with the same.

PR #1893, opened 2020-10-04
disambiguate columns used in query ordering

### LINKED ISSUE #1892: Cookbook API calls with user and order parameters results in error
# Describe the problem

Did you remove the ability to query the most downloaded cookbooks per user via the API? Whenever I try it, I get an app error. It only seems to work when you don't include the user parameter.

(Additional information from @robbkidd below) 

## Software Version

This was against the public Supermarket at supermarket.chef.io, then running 3.3.35.

## Replication Case

An HTTP GET to `https://supermarket.chef.io/api/v1/cookbooks?&user=<anyusername>&order=<anyorderoption>`. The `order` parameter could be any of [the cookbook ordering options](https://github.com/chef/supermarket/blob/3.3.35/src/supermarket/app/models/cookbook.rb#L29-L33).

## Stacktrace

On the client side, the client would receive a 500 error when the parameters `order` _and_ `user` are both present. 

<details><summary>PG::AmbiguousColumn: ERROR</summary>

```
ActionView::Template::Error: PG::AmbiguousColumn: ERROR:  column reference "id" is ambiguous
LINE 1: SELECT  DISTINCT cookbook_followers_count, id AS alias_0, "c...
                                                   ^
```
</details>

The query's inclusion of the user table to scope the results to a single user makes the column references in the ordering clauses ambiguous about which table's `id` (or `name` or `updated_at`) to use for ordering.

## Possible Solution

Include the `cookbooks` tablename in the column references in the order options SQL.

robbkidd:
## Description

Queries using the named ordering options and also joins with other tables were failing with:

    ActionView::Template::Error: PG::AmbiguousColumn:
    ERROR:  column reference "id" is ambiguous
    LINE 1: SELECT DISTINCT cookbook_followers_count, id AS alias_0, "c…

Database query was traced to these ordering options in the Cookbook scopes.

Added tests for the API route and query parameters that triggered these errors. Added tablenames to the column references to pass the tests.

## Related Issue

Fixes #1892

## Types of changes

- ~New feature (non-breaking change which adds functionality)~
- ~Breaking change (fix or feature that would cause existing functionality to change)~
- ~Chore (non-breaking change that does not add functionality or fix an issue)~

## Checklist:

- ~I have updated the documentation accordingly.~

### DIFF
--- a/src/supermarket/app/models/cookbook.rb
+++ b/src/supermarket/app/models/cookbook.rb
@@ -26,11 +26,11 @@ class Cookbook < ApplicationRecord
   }
 
   ORDER_OPTIONS = {
-    "recently_updated" => Arel.sql("updated_at DESC"),
-    "recently_added" => Arel.sql("id DESC"),
-    "most_downloaded" => Arel.sql("(cookbooks.web_download_count + cookbooks.api_download_count) DESC, id ASC"),
-    "most_followed" => Arel.sql("cookbook_followers_count DESC, id ASC"),
-    "by_name" => Arel.sql("name ASC"),
+    "recently_updated" => Arel.sql("cookbooks.updated_at DESC"),
+    "recently_added" => Arel.sql("cookbooks.id DESC"),
+    "most_downloaded" => Arel.sql("(cookbooks.web_download_count + cookbooks.api_download_count) DESC, cookbooks.id ASC"),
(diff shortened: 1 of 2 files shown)


davymcaleer (approved):
👍

PR #1906, opened 2021-02-22
bump openssl-1.0.2y

dheerajd-msys:
Signed-off-by: dheerajd-msys <dheeraj.dubey@msystechnologies.com>

Issue : https://github.com/chef/supermarket/issues/1907


tas50:
The openssl bump here needs to only bump the omnibus-software dep. Otherwise it will most likely break other parts of the build

dheerajd-msys:
Updated, Thank you @tas50 !!

PR #2065, opened 2021-07-21
Don't install berks now that it's in omnibus-toolchain

tas50:
The berks install is causing failures since it deps on chef/chef and
pulls in the latest which only works on Ruby 2.7+


tas50:
https://buildkite.com/chef/chef-supermarket-master-omnibus-adhoc/builds/70#60e72ba7-4cfd-49b5-a518-16830ff41961

tas50:
Still failing, but this got us around this issue

PR #2086, opened 2021-07-29
Switch back to compiled nodejs but use 12.22.3

tas50:
Binary nodejs fails the supermarket-ctl reconfigure. This should work
there, but it will give us a nice new version and hopefully npm/yarn
work with that new version.


tas50:
Better approach is being taken by the team

PR #2121, opened 2021-08-11
Remove bundler install in omnibus

tas50:
This is part of Ruby 2.7 so there's no need to isntall bundler again


tas50:
https://buildkite.com/chef/chef-supermarket-master-omnibus-adhoc/builds/139#52bf9f39-0c79-4aad-b46b-a3bbbf8fec93

PR #2122, opened 2021-08-11
Remove the binary nodejs

tas50:


tas50:
Works fine without this: https://buildkite.com/chef/chef-supermarket-master-omnibus-adhoc/builds/140

PR #2123, opened 2021-08-11
Switch from inspec to inspec-core

tas50:
This reduces our install size by over 100 megs by removing
azure/google/aws sdks that we're not actually using


tas50:
https://buildkite.com/chef/chef-supermarket-master-omnibus-adhoc/builds/141#7f00527b-76d3-41c8-9909-9ebaff768691

PR #2130, opened 2021-08-12
Add a cleanup definition for all the gem cache junk

tas50:
This should drop the install size by 70 megs


tas50:
https://buildkite.com/chef/chef-supermarket-master-omnibus-adhoc/builds/149#63d58a2a-25f3-4607-b99c-cc63b36261ed

PR #2166, opened 2021-08-18
Automate release announcements + minor expeditor config updates

tas50:
Pretty sure we need to modernize this whole config.

@tduffield Is this the old config format that needs to get reworked?


tduffield (approved):
Expeditor config looks good!

tas50:
Thanks for the quick response @tduffield. One more automated repo!

PR #2334, opened 2021-11-05
Allows dots in cookbook names

zedtux:
### Description

Creating a cookbook with a dot in the name is totally fine, and having a dot in a URL on the web is totally fine too. This commit allows dots in name in the supermarket.

A very good example of a usecase for this is the netplan packages in Ubuntu: There are two packages, one named 'netplan', and another one named 'netplan.io'. Both aren't addressing the same purpose so a dedicated cookbook for each of those packages is required. Reflecting the package names in the supermarket will avoid confusions to the users

With this commit, a cookbook URL like https://supermarket.chef.io/cookbooks/netplan.io will be allowed.

### Issues Resolved

None

### Check List

- [x] New functionality includes tests
- [x] PR title is a worthy inclusion in the CHANGELOG

### DIFF
--- a/src/supermarket/app/models/cookbook.rb
+++ b/src/supermarket/app/models/cookbook.rb
@@ -115,7 +115,7 @@ class Cookbook < ApplicationRecord
 
   # Validations
   # --------------------
-  validates :name, presence: true, uniqueness: { case_sensitive: false }, format: /\A[\w_-]+\z/i # rubocop:disable Rails/UniqueValidationWithoutIndex
+  validates :name, presence: true, uniqueness: { case_sensitive: false }, format: /\A[\w\._-]+\z/i # rubocop:disable Rails/UniqueValidationWithoutIndex
   validates :lowercase_name, presence: true, uniqueness: true
   validates :cookbook_versions, presence: true
   validates :source_url, url: { allow_blank: true }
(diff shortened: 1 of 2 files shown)


zedtux:
Looks like the failing job is not from my code.

zedtux:
Code rebased on top of `main`.

Stromweld:
I feel like this could potentially lead to future confusion or potential exploit as usually methods are defined after a period. Even if it's allowed in the supermarket, has it been tested to not create issues with chef-client runs itself?

RajeshPaul38 (dismissed):

zedtux:
Is there anything I could do in order to help this PR to be merged please?

lamont-granquist:
I'm against accepting this PR since it can be confused with method syntax, and could at least be confusing with ohai and knife search syntax where '.' is a record separator (this also affects IP addresses, but as there's not been any solution proposed for years, I'm not inclined towards making the problem there worse).

zedtux:
Well, if that the case, then an update of chef is required in order to prevent from installing cookbooks with a dot in their name since it is possible today (actually only the supermarket is forbidding it).
In my eyes you can't reject this PR if chef allows installing cookbooks with a dot in their name, both should be aligned.

Something that would fix all those potential issues (you're using "could", it does mean "to be check", doesn't it?) is that a cookbook [tells its name](https://gitlab.com/pharmony/netplan.io-cookbook/-/blob/master/metadata.rb#L3) and could be used in order to skip that first dot.
That needs to be verified and tested as I just though about it now.

saghoshprogress:
Closing this one, as we can't allow dot(.) in cookbook names

zedtux:
Well, that's really sad ... just for a little dot 🤷‍♂️. Whatever.

Issue #2285, opened 2021-10-04
Deprecating a cookbook should not make it show up as updated


tas50:
The updated cookbook list at https://supermarket.chef.io/cookbooks?order=recently_updated contains cookbooks that we deprecated. They should not be part of that list.

PR #2370, opened 2021-11-12
Filter deprecated cookbooks from cookbooks list

manick-vel-11:
### Description

This change is related to filtering out deprecated cookbooks from cookbooks list. Previously, both deprecated and undeprecated ones will be displayed, now deprecated ones will be removed from the list.

### Issues Resolved

https://github.com/chef/supermarket/issues/2285

### Check List

- [x] New functionality includes tests
- [x] All buildkite tests pass
- [x] Full omnibus build and tests in buildkite pass


manick-vel-11:
All cookbooks:
all_cookbooks_dep_undep


Removing deprecated cookbooks:
undep_cookbooks

dheerajd-msys:
Looks good!! Could you please attach adhoc build @manick-vel-11 ?

msys-sgarg:
Looks good manick

tas50:
I think we're going to need to make it so that users can still find these deprecated cookbooks if they choose. otherwise, they're just lost forever. You'll want to want to add a search option for this.

Something like this where the deprecated cookbook option is unchecked by default

image

src/supermarket/app/controllers/cookbooks_controller.rb
```diff
@@ -31,7 +31,12 @@ def index
       @cookbooks = @cookbooks.featured
     end
 
-    if @current_params[:order].present?
+    if (order = @current_params[:order]).present?
+
+      if order == "recently_updated"
+        @cookbooks = @cookbooks.undeprecated
```
btm, line 37: "undeprecated" sounds wrong. A cookbook is either deprecated, or it is not deprecated.

Can we `@cookbooks = @cookbooks - @cookbooks.deprecated` ?

Or change the scope to something like: `scope :not_deprecated, -> { where.not(id: deprecated) }` ?

msys-sgarg:
Please add the view specs also

manick-vel-11:
By default, only the not deprecated cookbooks is shown.
not_deprecated_cookbooks


If include deprecated cookbooks is checked, both the deprecated and not deprecated cookbooks will be shown.
include_deprecated_cookbooks

saghoshprogress (approved):

dheerajd-msys (approved):

PR #2375, opened 2021-11-15
Replace virtus gem with dry-types gem

### LINKED ISSUE #2349: Replace virtus gem with alternatives
The virtus gem is at it's EOL. So we need to replace it with a suitable alternative which has active contributions.
The alternatives can be found in the github page of virtus itself: https://github.com/solnic/virtus

Just for note the alternatives:
https://github.com/dry-rb/dry-types
https://github.com/dry-rb/dry-struct
https://github.com/dry-rb/dry-schema

RajeshPaul38:
### Description

Replace virtus gem with dry-types gem

### Issues Resolved

#2349 

### Check List

- [X] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/main/CONTRIBUTING.md#developer-certification-of-origin-dco>


src/supermarket/Gemfile
```diff
@@ -50,7 +50,7 @@ gem "unicorn"
 gem "unicorn-rails"
 gem "utf8-cleaner"
 gem "validate_url"
-gem "virtus", "1.0.2", require: false # Pin to a version before the handling of nil in collection coercion was fixed.
+gem "dry-struct", "~>1.4" # This gem is to replace the deprecated gem: virtus. Once tested virtus needs to be removed.
```
tas50, line 53: Since virtus is removed in this PR should this comment go away?

RajeshPaul38: sure

RajeshPaul38:
@tas50 any idea why is this failing:
**netlify/chef-supermarket/deploy-preview**

tas50:
@IanMadd Looks like netlify is not happy

IanMadd:
Rebase. This branches from a commit before the Netlify config/docs were added.

saghoshprogress:
Add a build link

RajeshPaul38:
In Progress Build: https://buildkite.com/chef/chef-supermarket-main-omnibus-adhoc/builds/276

> Add a build link

RajeshPaul38:
Build is successful @saghoshprogress

saghoshprogress (approved):

src/supermarket/app/lib/schema_definer/symbolize_struct.rb
```diff
@@ -0,0 +1,7 @@
+module SchemaDefiner
+  class SymbolizeStruct < Dry::Struct
```
msys-sgarg, line 2: Why do we need two modules ? Can this be combined to one module instead ? there is just one method and one class here ..  can we implement this class in types.rb also ?

RajeshPaul38: Actually there is need for one module and one class. In the current implementation also there are 2.

The class is `Dry::Struct` but along with that class I wanted to add certain behaviours i.e. `transform_keys(&:to_sym)`. For this reason I created a class as an interface in which you can attach different behaviours to be used by many classes (the implementations) and you can scale it the way you want.

But the module is a specific component of the dry-types gem which needs to be kept separate for need basis. But yes they could be part of the same file.

PR #2475, opened 2021-12-29
Update the announcement script for new downloads site

tas50:
Make sure we announce with the latest URL

### DIFF
--- a/.expeditor/announce-release.sh
+++ b/.expeditor/announce-release.sh
@@ -14,7 +14,7 @@ $(cat release-notes.md)
 ---
 ## Get the Build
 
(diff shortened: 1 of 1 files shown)


.expeditor/announce-release.sh
```diff
@@ -14,7 +14,7 @@ $(cat release-notes.md)
 ---
 ## Get the Build
 
-You can download binaries directly from [downloads.chef.io](https://downloads.chef.io/tools/supermarket/stable?v=$EXPEDITOR_VERSION).
+You can download binaries directly from [chef.io/downloads](https://www.chef.io/downloads/tools/supermarket?v=$EXPEDITOR_VERSION).
```
RajeshPaul38, line 17: when would this new link be available?

tas50: This new URL format is live now

PR #2495, opened 2022-01-05
Fixed github enterprise condition check.

antima-gupta:
Signed-off-by: antima-gupta <agupta@msystechnologies.com>

### Description

- Fixed GITHUB Enterprise conditions to display GITHUB account as default.

### Issues Resolved

[List any existing issues this PR resolves, or any Discourse or
StackOverflow discussions that are relevant]

### Check List

### DIFF
--- a/src/supermarket/app/helpers/application_helper.rb
+++ b/src/supermarket/app/helpers/application_helper.rb
@@ -48,7 +48,7 @@ module ApplicationHelper
   # @return [String]
   #
 
-  def github_eneterprise_enabled?
+  def github_enterprise_enabled?
     ENV["GITHUB_ENTERPRISE_URL"].present? &&
       (ENV["GITHUB_ENTERPRISE_URL"] != "YOUR_GITHUB_ENTERPRISE_URL") &&
       ENV["GITHUB_CLIENT_OPTION_SITE"].present? &&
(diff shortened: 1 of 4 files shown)


msys-sgarg (commented):
Suggest to put these static strings - YOUR_GITHUB_ENTERPRISE_URL in a Constant array sort of structure and check for the presence of value there .. That will look much cleaner according to me

antima-gupta:
> Suggest to put these static strings - YOUR_GITHUB_ENTERPRISE_URL in a Constant array sort of structure and check for the presence of value there .. That will look much cleaner according to me

@msys-sgarg we need to check each values should be present so still we have to check for all key's  values.

msys-sgarg:
> > Suggest to put these static strings - YOUR_GITHUB_ENTERPRISE_URL in a Constant array sort of structure and check for the presence of value there .. That will look much cleaner according to me
> 
> @msys-sgarg we need to check each values should be present so still we have to check for all key's values.

In that case may be maintain multiple constants .. this repeating of string at multiple places is problematic and can cause inconsistency in future

antima-gupta:
> > > Suggest to put these static strings - YOUR_GITHUB_ENTERPRISE_URL in a Constant array sort of structure and check for the presence of value there .. That will look much cleaner according to me
> > 
> > 
> > @msys-sgarg we need to check each values should be present so still we have to check for all key's values.
> 
> In that case may be maintain multiple constants .. this repeating of string at multiple places is problematic and can cause inconsistency in future

Repeating of string is only on two places , we could make it at one place if it would not be use in `initializers/omniauth.rb`

msys-sgarg (approved):
Changes look good to me

saghoshprogress (approved):

PR #2514, opened 2022-01-10
Use the new PNG supermarket image

tas50:
New image that matches the site's appearance in 2022.


kagarmoe (approved):

kagarmoe (commented):
Doesn't render (yet).

kagarmoe (changes requested):
We need to make sure it renders before merging.

kagarmoe:
...And then I merged the wrong one first...

PR #2569, opened 2022-02-02
Disable Host Header Poisoning against Chef Supermarket

RajeshPaul38:
### Description

Supermarket currently allows arbitrary Host header. As a result all the relative URLs in the rendered page will have the passed Host header as a prefix which will result in the URLs belonging to the passed Host header. We need to stop this so that the current domain name of supermarket is considered as the domain for the relative URLs and thus the links will work properly.

### Issues Resolved

INFSUS-43

### Check List

- [X] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/main/CONTRIBUTING.md#developer-certification-of-origin-dco>
- [X] PR title is a worthy inclusion in the CHANGELOG

### DIFF
--- a/src/supermarket/config/environments/test.rb
+++ b/src/supermarket/config/environments/test.rb
@@ -43,5 +43,4 @@ Rails.application.configure do
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
-
 end


RajeshPaul38:
Build: https://buildkite.com/chef/chef-supermarket-main-omnibus-adhoc/builds/339

antima-gupta:
Looks good to me.

RajeshPaul38:
Checking main build: https://buildkite.com/chef/chef-supermarket-main-omnibus-adhoc/builds/342

RajeshPaul38:
> unit test are failing @RajeshPaul38 please check.

Checking in local what exactly is failing @antima-gupta

RajeshPaul38:
> unit test are failing @RajeshPaul38 please check.

Unit test failures have been fixed.

RajeshPaul38:
New build: https://buildkite.com/chef/chef-supermarket-main-omnibus-adhoc/builds/343

RajeshPaul38:
Rebased with main.
Build link: https://buildkite.com/chef/chef-supermarket-main-omnibus-adhoc/builds/348
In progresss

RajeshPaul38:
Buildkite build is successful: https://buildkite.com/chef/chef-supermarket-main-omnibus-adhoc/builds/348

antima-gupta (approved):
@RajeshPaul38 please checkout this file src/supermarket/config/environments/test.rb

src/supermarket/config/environments/test.rb
```diff
@@ -43,4 +43,5 @@
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
```
antima-gupta, line 46: No change in this file.

antima-gupta (changes requested):

RajeshPaul38:
@tas50 can you please review this? Do we actually need to update the documentation for this change?

omnibus/cookbooks/omnibus-supermarket/attributes/default.rb
```diff
@@ -359,6 +359,10 @@
 # recommend false; set to true as default for backward compatibility
 default['supermarket']['owners_can_remove_artifacts'] = true
 
+# Setting allowed host for supermarket to avoid arbitrary "Host" header injection
+# Set this value to the domain name of your supermarket website e.g. supermarket.chef.io
+default['supermarket']['allowed_host'] = node['supermarket']['fqdn']
```
tas50, line 364: What does this allow specifically and what's the impact to customers? Will people need to change this?

RajeshPaul38: Basically this new configuration will help the customers of supermarket prevent any host header injection attack which is a common CVE now a days. Basically with host header injection it's possible to lead the supermarket users to a completely different domain. To prevent this  we have introduced this attribute where the customer can whitelist the hostname of supermarket in the list of allowed hosts. This way any random domain names passed in as Host header in the request to supermarket will be completely discarded thus avoiding such attacks.

So by default this will be set as the FQDN/hostname of the instance where supermarket will be running. In case the user has a custom domain in which supermarket will be running the user simply needs to set the FQDN in the `/etc/supermarket/supermarket.rb` as the intended domain name of supermarket and thus the `allowed_host` will be set accordingly with the command: `supermarket-ctl reconfigure`. So he doesn't need to explicitly set this value.

antima-gupta (approved):

msys-sgarg (approved):
Looks good to me

RajeshPaul38:
@tas50 We are awaiting your feedback. Otherwise we are not able to close this.

pratixha (approved):

tas50:
This is probably fine if there's no way to avoid it but we will need to make sure all the docs are updated with this new step. For most users the FQDN will not be the actual supermarket host so they will need to set this.

RajeshPaul38:
@tas50 in the docs site the FQDN setup is mentioned already. So it's already part of the setup steps. So we don't need any further changes in the doc specific to this change. So are we good to merge this?
Screenshot 2022-02-16 at 4 32 58 PM

RajeshPaul38:
This is to fix this customer issue: https://github.com/chef/customer-bugs/issues/415

PR #2574, opened 2022-02-04
Pull in the latest cookstyle and allow dependabot to bump it

tas50:
There are lots of cookstyle releases and we want to make sure we can always pull in new releases.


dheerajd-msys:
DCO sign-off is missing @tas50

tas50:
@dheerajd-msys DCO is there. The bot just isn't the best

dheerajd-msys:
Adhoc build https://buildkite.com/chef/chef-supermarket-main-omnibus-adhoc/builds/351 is in progress. Can approve post it's success

src/supermarket/Gemfile
```diff
@@ -82,7 +82,7 @@ group :test do
 end
 
 group :development, :test do
-  gem "chefstyle", "2.1.3"
+  gem "chefstyle"
```
antima-gupta, line 85: why are we using this `cookstyle` gem in common gems as well as inside group?

dheerajd-msys (approved):

src/supermarket/Gemfile
tas50: This is chefstyle which is used for our ruby code vs. cookstyle which is used for cookbook code

PR #2584, opened 2022-02-07
Remove requirements since these are bundled

tas50:
It doesn't make much sense to call these out as requirements since this is all packages. If we want to call out the currently packaged tools and what you would need for development that should live elsewhere.


msys-sgarg (approved):
looks good to me

PR #3406, opened 2026-07-14
Remove stale brakeman gem dependency

RoyShravani:
## Summary

Removes the stale `brakeman` gem entry from `Gemfile` and `Gemfile.lock`.

## Why

- The `brakeman 5.4.1` entry in `group :development` was never actively used
- The CI workflow (`.github/workflows/brakeman-analysis.yml`) independently installs its own pinned version (`5.0.4`) via `gem install` and does not reference the Gemfile at all
- The gem is excluded from the omnibus package build via `bundle config set without 'development doc'`, so it was never distributed
- The 5.4.1 entry was being picked up by BlackDuck/dependency scans, raising licensing questions around the Brakeman Public Use License unnecessarily

## What is unaffected

- The `brakeman-analysis.yml` security scan workflow continues to run unchanged — it installs Brakeman 5.0.4 independently on GitHub-hosted runners
- No application code, tests, or CI pipelines reference the Gemfile entry

## Changes

- `src/supermarket/Gemfile` — removed `gem "brakeman"` from `group :development`
- `src/supermarket/Gemfile.lock` — removed corresponding `brakeman (5.4.1)` entries


RoyShravani:
adhoc build: https://buildkite.com/chef/chef-supermarket-main-omnibus-adhoc/builds/823

saghoshprogress (approved):

